PR #60, opened 2016-03-07
Blockstore --> Blockstack

jcnelson:
Just a rename; should also be safe to merge to master.  I've updated blockstack-client in its `develop` branch.


muneeb-ali:
I copied a bunch of these changes over to the latest version. The registrar is now part of CLI. Closing the pull request.

PR #76, opened 2017-10-17
Add headerless proof checks

kantai:
This checks social proofs if configured to do so by reading them directly out of the user's 0th profile.


kantai:
Merging support for proof checks via reading the profile.json directly.

PR #97, opened 2018-03-12
Add Heroku One Click deploy button

jackzampolin:
This PR adds a 1 click deploy button for Heroku. The one remaining issue is that the servername is not available in heroku (app name) and is either user generated or randomly generated. This means that the `servername` configuration does not change the server name in the `/hub_info` response.


kantai:
This works great @jackzampolin!

I ended up using the heroku "secret" generator to set the server name portion of the challenge text --- that ensures that tokens signed for one gaia hub won't be usable on a different gaia hub.

PR #100, opened 2018-03-28
Add a disk storage driver

jcnelson:
This pull request does a couple things (happy to send them separately!)

* Removes dead Python storage drivers and setup.py script, which are waaaaaay out of sync from blockstack-core anyway (and they're about to be deprecated in favor of Gaia hubs)
* Removes old Gaia figures that no longer characterize the way this system behaves
* Adds a disk driver, including tests and a sample configuration
* Makes it so `npm install -g` installs a `blockstack-gaia-hub` program


kantai:
This looks great @jcnelson and it works when I tested it on my machine! You can merge this PR whenever you're ready.

jcnelson:
Thanks! Merging...

Issue #114, opened 2018-05-15
Gaia authentication payload must include client-side data


kantai:
The authentication process for authenticating gaia hub should include client-generated content in the signed payload (and will therefore need a way to relay that data to the gaia hub as well). This would prevent a malicious gaia hub from requesting arbitrary challenge texts, because the client would be signing a payload with client-generated data, in addition to the challenge text.

kantai:
This was addressed in #115, and is released in the latest version of the gaia hub.

PR #115, opened 2018-05-17
Authentication with JWTs

kantai:
This adds support for authentication using ES256K JWTs (using the `jsontokens` library).

Such auth strings must be sent with a `v1:` prefix, which allows for backwards compatibility with the current authentication scheme.

This authentication scheme allows clients to add other fields to their JWTs (e.g., a `salt`) or an expiration time (the `exp` [claim](https://openid.net/specs/draft-jones-json-web-token-07.html#ReservedClaimName)).

This addresses #114 

Additionally, this PR does two other things:

1. Explicitly sets the node version to 8 -- this is the current LTS. This insures that the node version will match in the Dockerfile, the circleCI setup, and package.json
2. Adds `proxyquire` mocks to the unit tests if a configuration file isn't explicitly set for a given driver.


hub/src/server/authentication.js
```diff
@@ -0,0 +1,191 @@
...
+  constructor(token: string) {
+    this.token = token
+  }
+
+  static fromAuthPart(authPart: string) {
+    if (!authPart.startsWith('v1:')) {
+      throw new ValidationError('Authorization header should start with v1:')
+    }
+    const token = authPart.slice('v1:'.length)
+    const decodedToken = decodeToken(token)
```
jcnelson: Consider surrounding this with a try/catch block that throws ValidationError(), since the token could be malformed.

hub/src/server/authentication.js
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  isAuthenticationValid(address: string, challengeText: string,
+                        options?: { throwOnFailure?: boolean }) {
+    const defaults = {
+      throwOnFailure: true
+    }
+    options = Object.assign({}, defaults, options)
+
+    const decodedToken = decodeToken(this.token)
```
jcnelson: Consider surrounding with a try/catch block that throws ValidationError if we couldn't decode the token.

hub/src/server/authentication.js
```diff
@@ -0,0 +1,191 @@
...
+    const publicKey = decodedToken.payload.iss
+    const gaiaChallenge = decodedToken.payload.gaiaChallenge
+
+    try {
+      if (! publicKey) {
+        throw new ValidationError('Must provide `iss` claim in JWT.')
+      }
+
+      const issuerAddress = pubkeyHexToECPair(publicKey).getAddress()
```
jcnelson: Do we mandate that the public key always be compressed?  If so, we could enforce this check as a separate error message (i.e. by checking the public key length), and throwing a validation error saying that the public key format is invalid (or converting it ourselves and logging a warning).

hub/src/server/authentication.js
```diff
@@ -0,0 +1,191 @@
...
+      if (issuerAddress !== address) {
+        throw new ValidationError('Address not allowed to write on this path')
+      }
+
+      const verified = new TokenVerifier('ES256K', publicKey).verify(this.token)
+      if (!verified) {
+        throw new ValidationError('Failed to verify supplied authentication JWT')
+      }
+
+      if (gaiaChallenge !== challengeText) {
```
jcnelson: I think this check should happen earlier than the signature verification, so we don't put ourselves in a position where the attacker can observe a victim decrypt invalid data.

hub/src/server/authentication.js
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  isAuthenticationValid(address: string, challengeText: string,
+                        options?: { throwOnFailure?: boolean }) {
+    const defaults = {
+      throwOnFailure: true
+    }
+    options = Object.assign({}, defaults, options)
+
+    if (this.publickey.getAddress() !== address) {
```
jcnelson: Same question about the public key being compressed/uncompressed applies here.

hub/src/server/authentication.js
```diff
@@ -0,0 +1,191 @@
...
+
+    if (options.throwOnFailure && !valid) {
+      logger.debug(`Failed to validate with challenge text: ${challengeText}`)
+      throw new ValidationError('Invalid signature or expired authentication token.')
+    }
+    return valid
+  }
+}
+
+export function getChallengeText(myURL: string = DEFAULT_STORAGE_URL) {
```
jcnelson: Does this method mandate a minimum challenge text length?

hub/src/server/http.js
```diff
@@ -78,10 +81,11 @@ export function makeHttpServer(config: Object) {
 
   app.get('/hub_info/', (req: express.request,
                          res: express.response) => {
-    const challengeText = StorageAuthentication.challengeText(server.serverName)
+    const challengeText = getChallengeText(server.serverName)
     const readURLPrefix = driver.getReadURLPrefix()
-    writeResponse(res, { 'challenge_text' : challengeText,
-                         'read_url_prefix' : readURLPrefix }, 200)
+    writeResponse(res, { 'challenge_text': challengeText,
```
jcnelson: We should consider not replying anything if the challenge text is too short (i.e. give back a 400).

hub/src/server/authentication.js
kantai: No, we don't currently -- but I suppose we could. I think if we wanted to go that route, the rest of our tooling should only use compressed public keys.

hub/src/server/authentication.js
kantai: Not sure what the leak would be here -- nothing is being decrypted in the authentication process.

jcnelson: Yup.  Updated comment.

hub/src/server/http.js
kantai: Hmm -- I like that approach, will add.

kantai:
@jcnelson -- If have you have a chance, can you take a look at the changes and responses since your review? Thanks!

jcnelson:
Looks good to me!

PR #120, opened 2018-06-11
Consistent readurl config setting

kantai:
This moves the logic of the readURL setting out of the drivers. Work still needs to be done to make the disk driver consistent as well.


kantai:
This works with the file driver as well. Merging into develop.

Issue #79, opened 2017-12-14
Hub: Add IPFS Driver


kantai:
We should support an IPFS backend for the Gaia Hub.

However, we should probably pin down a spec for gaia hub drivers _before_ we do this.

bornbrie:
This would be an amazing addition!

koalalorenzo:
This would be AWSOME :D Any plan?

zone117x:
PR https://github.com/blockstack/gaia/pull/129 for this which needs some updating and tests.

@kantai I'm currently doing work on fixing up our Gaia tests. I'd be interested in taking this on - pending some clarity on current priorities. Thoughts?

zone117x:
Closing this issue due to lack of activity. Please feel free to comment if this should be re-opened.

tesla809:
Considering the advancements of IPFS and its ecosystem and tools, it would be wise to add support.

Could we re-open this issue?

cooladam201212:
Yes, agree.
To me, blockstack seems like an AWS thing in blockchain area.
So it should keep integrating the best decentralized components into it.

criadoperez:
PR #427

PR #125, opened 2018-06-27
#79 Add IPFS Driver

dsdjolence:
This pull request will allow users to use IPFS as a storage. Related to issue #79.

You have to run your own IPFS node, also check that your node supports API or enable it:

**Show the IPFS config API port to check it is correct:**
`$ ipfs config Addresses.API`
_output: /ip4/127.0.0.1/tcp/5001_

**Set it if it does not match the above output:**
`ipfs config Addresses.API /ip4/127.0.0.1/tcp/5001`
_Restart the daemon after changing the config._

**Run daemon:**
`ipfs daemon --enable-namesys-pubsub`

`--enable-namesys-pubsub` is a flag and it will enables pubsub and this flag isn't required. For more details check IPFS [documentation](https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md#ipns-pubsub).

In order to enable this driver edit your `config.json`:
```
"driver": "ipfs",
...

"ipfs": {
  "ipnsPath": <required (e.g: "blockstack")>,
  "readProviderUrl": <optional (default: "https://gateway.ipfs.io")>,
  "peerKeyName": <optional (default: "self")> ,
  "configuration": {
    "host": <required (e.g: "localhost")>,
    "port": <required (e.g: "5001")>,
    "protocol": <required (e.g: "http")>,
    "api-path": "<optional>"
  }
},
```


kantai:
Thanks for this @dsdjolence !

It looks like you're using IPNS, which makes sense, and matches with roughly what we figured a IPFS driver would require. I should be able to review this sometime this weekend, and I'll try to rebase this to `develop` as well, because this driver doesn't need a read-path.

dsdjolence:
Thanks @kantai for the feedback. Because this driver doesn't need a read-path I have created pull request on wrong branch, well if you need I can fix it.

kantai:
Merging this to a branch I can edit, and then will open another PR with some changes to remove the read-url components.

Issue #96, opened 2018-03-12
Add support for application addresses with whitelisting


kantai:
This has to do with governing which addresses a gaia hub will accept writes on behalf of, which is a necessity for operating private hubs (https://github.com/blockstack/gaia#private-hubs)

Currently, the `config.json` file governs address whitelisting in Gaia. If a user wishes to restrict the set of addresses which can write to the gaia hub, this can pose a major usability issue.

Basically, if you want to whitelist an "identity address" like `aaron.id`'s `1EJh2y3xKUwFjJ8v29a2NRruPJ71neozEE` -- you cannot _actually_ derive, say, my application address for app `foo` from the ID (or even the `xpub`, because these are all bip32 hardened indexes)

There's a couple of approaches we could take:

1. Add endpoint to the gaia hub which will allow the blockstack-browser to modify the whitelist. This would require the notion of "admin addresses" which can modify the whitelist (you wouldn't want applications, for example, to have this access)

2. `blockstack.js` attaches a "proof" in the authentication header on gaia hub writes. This proof would show that the app address is derived from the identity address. The auth response token _already_ proves this fact (by signing the JWT with the identity key, and including the app private key) --- however, this would require (1) sharing the auth response token and (2) the app's transit key, which we should _not_ do. Instead, we'd probably want to include this as a new field in the auth response as a JWT with `payload = {  appPublicKey, identityPublicKey }` and `signer = identityPublicKey`

I'm in favor of (2), as that's probably the simplest to implement, and it keeps the gaia hub basically stateless (you would just whitelist _identity addresses_, and then the application addresses for those identities would follow from there.)

Would love to hear thoughts on this -- tagging some people who I remember having views on this: @jcnelson @jackzampolin @larrysalibra

jackzampolin:
I agree that 2 would be far and away easiest to implement, however I think an endpoint to manage the configuration would be a really nice feature that we would likely end up building out in the future.

This feature would be a nice enabler for hosted gaia hubs (the service would want to automatically adjust the whitelist based on an outside DB)

kantai:
> This feature would be a nice enabler for hosted gaia hubs (the service would want to automatically adjust the whitelist based on an outside DB)

Outside DBs for the Gaia hub are exactly the kind of thing I think we should try really *really* hard to avoid -- because this basically means that we now have this additional piece of user-state which is neither really gaia-storage data, nor data that lives on-chain, and then that means we'll need additional processes around migration and recovery there.

[...]

PR #135, opened 2018-07-25
Feature/associationtoken

jcnelson:
This implements the `associationToken` scheme (#96), which allows a user to sign an "inner" JWT with a whitelisted key that authorizes another key (e.g. an application) to write to the Gaia hub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,9 +114,11 @@ structure is:
 {
  'type': 'object',
  'properties': {
-   'iss': { 'type': 'string' }
-   'exp': { 'type': 'IntDate' }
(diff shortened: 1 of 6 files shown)


kantai:
This looks great to me -- I updated it to pass the flow tests (to make the functions consistently return strings, I removed the old optional parameter, `throwOnFailure`, which was only ever used in testing).

Can you update the documentation here https://github.com/blockstack/gaia/blob/master/README.md#v1-authentication-scheme ?

jcnelson:
Just pushed.  Thanks for reviewing!

Issue #127, opened 2018-06-29
Add check to authentication which requires the URL to be communicated


kantai:
We can add a requirement to the authentication which ensures that the authentication token was generated by a client intending to communicate with this gaia hub -- basically, just add a claim like:

```
hubUrl: "https://hub.blockstack.org"
```

And then the Gaia hub asserts that the hubURL in the authentication token matches the one that the hub expects of itself.

jcnelson:
Shipped!

PR #138, opened 2018-07-31
Authentication with checked hubUrl

kantai:
This addresses #127 -- tests added for this validation.

This is a non-breaking change, as it's a configurable option that defaults to the current behavior.

### DIFF
--- a/hub/src/server/authentication.js
+++ b/hub/src/server/authentication.js
@@ -43,7 +43,7 @@ export class V1Authentication {
   }
 
   static makeAuthPart(secretKey: bitcoin.ECPair, challengeText: string,
-                     associationToken?: string) {
+                      associationToken?: string, hubUrl: string) {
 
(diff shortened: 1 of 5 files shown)


hub/src/server/authentication.js
```diff
@@ -259,7 +283,15 @@ export function parseAuthHeader(authHeader: string) {
 }
 
 export function validateAuthorizationHeader(authHeader: string, serverName: string,
-                                            address: string) : string {
+                                            address: string, requireCorrectHubUrl?: boolean = false,
+                                            validHubUrls?: ?Array<string> = null) : string {
+  const serverNameHubUrl = `https://${serverName}`
```
jcnelson: I think I know the answer, but just want to make sure---there's never going to be a world where we support `http`, right?  I don't think it would work in Web browsers anyway, unless the app itself was served via `http`.

kantai: Yep -- not a world that Gaia wants to live in.

PR #139, opened 2018-08-01
Fix list-files for disk

jcnelson:
Fix the list-files implementation for the disk driver so it lists all files (like `find`), not just only the top-level directories.  This makes it consistent with the rest of the drivers on list-files.  Update the tests to test this new behavior.


kantai:
Good catch @jcnelson -- merging to develop now.

Issue #142, opened 2018-08-28
Scoped association tokens


jcnelson:
A question I get a lot is how to allow one user to write to a *limited subset* of another user's Gaia hub.  The way this would be achieved would be to expand Gaia's `associationToken` to include scopes.  I'm thinking we'd add the following:

* Scope writes to a particular prefix within the Gaia hub
* Allow writes but not list-files
* Allow list-files but not writes

Thoughts? @kantai

markmhx:
This sounds intriguing. I'd be curious to hear about the functional use cases behind the question to better understand the types of data developers would allow users to write to each others' hubs.

hstove:
@markmhx yes as Thomas said, this is exactly how I'd like to build out collaboration functionality in Kanstack.

I know it's maybe not a priority right now, but is there anything I can do to help move this along? I could definitely write some blockstack.js code to accommodate  this at some point.

jcnelson:
I updated the issue to say "authentication token," since this is what users ask for.  They want multiple users to be able write to the same *bucket* in the same Gaia hub, not just write to the same Gaia hub.  To clarify the high-level difference:

* association tokens are what allow you to write to a particular Gaia hub
* authentication tokens are what allow you to write to a specific bucket in a particular Gaia hub

markmhx:
Thanks for the clarifactions, @hstove and @jcnelson.

Can we also update the issue's title so that it's more imperative in nature, and as such describes to core task here. e.g. "Expand authentication token support to include scopes"?

If we already have authentication tokens but not these scopes, what are they currently used for? Allowing writes and list-files for all buckets in a given hub?

kantai:
> If we already have authentication tokens but not these scopes, what are they currently used for? Allowing writes and list-files for all buckets in a given hub?

The authentication tokens are used to allow writes and list-files for a specific bucket. Scoping would allow us to create tokens which only grant write access to, say, specific paths.

kantai:
Is it okay to not support list-files in the authentication token? I want to make sure that we're not encouraging a use case where applications are doing a lot of list-files queries when other strategies would work. If there's a compelling use case for non-owners to perform list-files, I'd be okay with that. But here's roughly what I'd like to see in a scoped auth token:

```
pathsAllowed: [ string ],
maxFileSize: number
```

What were you thinking for the pathsAllowed? Supporting wildcards, or just explicit paths?

hstove:
[...]

PR #145, opened 2018-09-16
Feature/scoped authtoken

jcnelson:
https://github.com/blockstack/gaia/issues/142

### DIFF
--- a/hub/src/server/authentication.js
+++ b/hub/src/server/authentication.js
@@ -15,6 +15,16 @@ function pubkeyHexToECPair (pubkeyHex) {
   return bitcoin.ECPair.fromPublicKeyBuffer(pkBuff)
 }
 
+export type AuthScopeType = {
+  scope: string,
+  domain: string
+}
+
+export const AuthScopes = [
+  'putFile',
+  'putFilePrefix'
+]
+
 export class V1Authentication {
   token: string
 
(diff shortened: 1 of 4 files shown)


hub/src/server/server.js
```diff
@@ -69,6 +68,48 @@ export class HubServer {
...
+      } else if (scopes[i].scope == 'putFile') {
+        writePaths.push(scopes[i].domain)
+      }
+    }
+
+    if (writePrefixes.length > 0 || writePaths.length > 0) {
+      // we're limited to a set of prefixes and paths.
+      // does the given path match any prefixes?
+      let match = false
+      if (writePrefixes.length > 0) {
```
kantai: This `if` statement is not necessary -- that next `for` loop won't run if writePrefixes is empty.

hub/src/server/authentication.js
```diff
@@ -306,3 +372,49 @@ export function validateAuthorizationHeader(authHeader: string, serverName: stri
...
+  if (scopes.length > 8) {
+    throw new ValidationError('Too many authentication scopes')
+  }
+
+  for (let i = 0; i < scopes.length; i++) {
+    const scope = scopes[i]
+    
+    // valid scope?
+    let valid = false
+    for (let j = 0; j < AuthScopes.length; j++) {
```
kantai: This code could be a bit simpler if you use `.find()` instead of this inner loop

hub/src/server/server.js
```diff
@@ -69,6 +68,48 @@ export class HubServer {
...
+          if (path.startsWith(writePrefixes[i])) {
+            match = true
+            break
+          }
+        }
+      }
+
+      if (!match) {
+        // check for exact paths 
+        for (let i = 0; i < writePaths.length; i++) {
```
kantai: You could use `.find()` or `.indexOf()` here instead of the loop/break

hub/src/server/server.js
```diff
@@ -69,6 +68,48 @@ export class HubServer {
...
+        writePaths.push(scopes[i].domain)
+      }
+    }
+
+    if (writePrefixes.length > 0 || writePaths.length > 0) {
+      // we're limited to a set of prefixes and paths.
+      // does the given path match any prefixes?
+      let match = false
+      if (writePrefixes.length > 0) {
+        for (let i = 0; i < writePrefixes.length; i++) {
```
kantai: Actually, same advice re: find applies here. I think using `.find` over loop/breaks improves the code clarity --- however, if you feel strongly otherwise, that's okay with me.

jcnelson:
I didn't know about `find()`.  I updated the code to use it instead.

PR #153, opened 2018-11-01
Add support for environment variable configuration

kantai:
This adds support for configuring the gaia hub using environment variables rather than a config.json file.

This adds a static method to the driver classes which will specify the driver's default config options, and load the environment variables for the driver if provided.

A list of environment variables is provided in the `envvars.md` file.

Additionally, this PR add support for the `Cache-Control` setting to the GCP and S3 drivers. This is particularly important for the GCP driver, because it defaults to a 1 hour cache setting.


hstove:
> Do we anticipate that the config file contents will get so big that this will be infeasible?

A quick google search indicates that this won't be a problem.

But it's definitely cumbersome. A benefit is that we don't have to ever make updates to match config file logic to ENV variable logic. It will 'just work'.

We could have some UI in the 'one step deploy' documentation that lets you input your config, and it would auto-generate the config string for you. We should probably do that in any case.

jcnelson:
My question stems from ambiguity over how users will interact with Gaia via environment variables.

On the one hand, if we expect the user to *programmatically* pass configuration state into a Gaia hub without writing it to a file (i.e. by using some other program, like an admin tool, to generate the config data and feeds it into a co-located Gaia hub), then having a single environ with the serialized configuration JSON makes more sense to me.  Let the other program figure out a user-friendly way to enter the data.

On the other hand, if we expect the user to *manually* pass configuration state to the Gaia hub, then having a lot of individual environs makes more sense.  A list of environs is easier to write and manage than a blob of JSON.

kantai:
For environment variables, I think the use case much more closely matches a developer or administrator manually configuring a deployment. And for the near term, that’s the only use case we have.

For a programattic use case, I think there’s no need for the environment variables at all, as the UI or what-have-you would likely just write the JSON file directly.

jcnelson:
Thank you for your feedback @robert-bo-davis, that was very helpful.  I'm now convinced that @kantai's current approach is the right approach.

jcnelson:
@kantai are there unit tests for loading up and checking the valid setting of these envars?

kantai:
@jcnelson -- no, in the past we haven't had unit tests for config setting reads, but we've definitely had bugs in other repos related to that in the past, so we probably should. Will add.

kantai:
Updated with unit tests for the config changes. Also changed the "overwriting" behavior slightly so that if you change a single item in a nested object in the config, it doesn't overwrite all config settings (i.e., do a deep merge rather than a shallow merge).

Also refactored the unit tests so that they are no longer in a single massive file.

jcnelson:
Looks good to me!  Thanks for refactoring the test.js file :+1:

PR #155, opened 2018-11-05
Hotfix/disk content type

jcnelson:
This patch updates the disk driver to store the `content-type` header contents alongside any files, written into the directory whose path is `$storageRoot/.gaia-metadata/$address/$filename`.  This is important, because these metadata files will *not* be directly listable or writeable by remote clients.

This PR also adds a `reader/` sub-package that implements the read endpoint for the `disk` driver.  This is a necessary component because the read endpoint is aware of the disk driver's above scheme, and will supply the appropriate content-type header on `GET`.

When deploying a Gaia hub that uses the `disk` driver, it is recommended to run both the `blockstack-gaia-hub` process and the `blockstack-gaia-reader` process, with the `blockstack-gaia-reader` process running behind a "real" HTTP server like nginx.  The `blockstack-gaia-reader` process can read the same config file as the `blockstack-gaia-hub`, and will honor the `diskSettings.storageRootDirectory` field to find the top-level disk storage directory.

### DIFF
--- a/reader/src/server.js
+++ b/reader/src/server.js
@@ -1,0 +1,33 @@
+/* @flow */
+
+import Path from 'path'
+import fs from 'fs'
+
+const METADATA_DIRNAME = '.gaia-metadata'
+
+export class GaiaDiskReader {
+
+  constructor() {
+  }
+
+  handleGet(storageRoot: string, topLevelDir: string, filename: string) : Promise<*> {
+    const filePath = Path.join(storageRoot, topLevelDir, filename)
+    try {
+      fs.statSync(filePath)
+    } catch (e) {
+      const ret = { exist: false, contentType: undefined }
+      return Promise.resolve().then(() => ret)
+    }
+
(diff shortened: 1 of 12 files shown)


reader/src/server.js
```diff
@@ -0,0 +1,33 @@
...
+import fs from 'fs'
+
+const METADATA_DIRNAME = '.gaia-metadata'
+
+export class GaiaDiskReader {
+
+  constructor() {
+  }
+
+  handleGet(storageRoot: string, topLevelDir: string, filename: string) : Promise<*> {
```
kantai: this should just be a function if there aren't any instance variables

kantai:
Pushed a simple error catch to the read endpoint. I was testing whether or not you can traverse with URLs like:

```
http://localhost:8008/1FmbX4xLUdf28RKGz7aFajvMTCvXmAnh8M/../../baz
```

You cannot, but it caused the `sendFile` to barf an uncaught promise rejection.

kantai:
Anyways, thanks for the changes and the fix @jcnelson, please bump the version in `package.json` and then feel free to merge.

PR #157, opened 2018-12-06
Fix reader bug with storageRootDirectory config key used for file GETs.

zone117x:
Tagging @GInxh since this may be affecting work on the Gaia VMs.


jcnelson:
Looks good to me :+1: Thanks!

PR #158, opened 2018-12-06
Fix disk driver bug with the recursive-mkdir function.

zone117x:
When specifying an absolute path like `/tmp/gaia` the function that ensures the directory exists (mimicking `mkdir -p`) was failing on MacOS. Looks like it was doing an lstat on the path `/` and ending up trying to create a path like `//tmp/gaia`. This may have been working on linux - not sure.

Replaced the function with the popular `fs-extra` lib which includes a cross platform mkdirp/ensureDir.

### DIFF
--- a/hub/package.json
+++ b/hub/package.json
@@ -17,6 +17,7 @@
     "cors": "^2.8.4",
     "express": "^4.15.4",
     "express-winston": "^2.4.0",
+    "fs-extra": "^7.0.1",
     "jsontokens": "^0.7.8",
     "node-fetch": "^2.0.0",
     "winston": "^2.3.1"
--- a/hub/src/server/drivers/diskDriver.js
+++ b/hub/src/server/drivers/diskDriver.js
@@ -53,7 +53,15 @@ class DiskDriver implements DriverModel {
 
   mkdirs(path: string) {
     const normalizedPath = Path.normalize(path)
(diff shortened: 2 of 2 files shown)


jcnelson:
I've only tested on Linux.  Thanks for discovering this!

Instead of pulling in a whole new library, can we just add a check that the path doesn't start with a double-`/`?

zone117x:
> Instead of pulling in a whole new library, can we just add a check that the path doesn't start with a double-`/`?

I'm definitely wary about npm dependency bloat as well but in this case I think it may be useful.

I saw the readme note -
> Note: The disk driver requires a *nix like filesystem interface, and will not work correctly when trying to run in a Windows environment.

What do you think about leaving in this dependency for helping with future cross platform support (mostly Windows)? The lib can help deal abstracting out pathing / filesystem intricacies.

In this mkdir instance the lib does a bit to help with that - https://github.com/jprichardson/node-fs-extra/blob/5b29ae3aa5198c33820a07eaa4493a5b6d01fa82/lib/mkdirs/mkdirs-sync.js#L9

kantai:
This looks good to me, and I think in this case, `fs-extra` will be useful. It's also a _very_ widely used library: https://www.npmjs.com/package/fs-extra (it's in the top 20 most depended on libraries).

PR #159, opened 2018-12-06
Update Babel & add out-of-the-box debug config for VSCode

zone117x:
* Update to new major version of Babel, and performing minimal transforming (just Flow type removal and es6->commonjs module transform).
* Added .vscode config files for easy debugging on any/all three projects. The debugging is conveniently done directly on Flow src/*.js files (this capability required the Babel and Flow updates).
* Minor update to gaia-reader to allow for the same storage config env param as gaia-hub. This allows the project to be pulled and ran/debug in vscode with no configuration file setup.

-----

This PR contains a couple hotfixes which where cherrypicked and PR'd to master - see the following PRs for discussion on those changes:
* https://github.com/blockstack/gaia/pull/158
* https://github.com/blockstack/gaia/pull/157


admin/src/server.js
```diff
@@ -33,7 +31,8 @@ function runSubprocess(cmd: string, argv: Array<string>, env: Object, uid: ?numb
       opts.setgid = gid
     }
    
-    return new Promise((resolve, reject) => {
+    return new Promise((resolve) => {
+      /* eslint-disable-next-line camelcase */
       child_process.spawn(cmd, argv, opts)
```
jcnelson: Let's just change this to `childProcess`.

hub/src/server/drivers/AzDriver.js
```diff
@@ -69,7 +69,7 @@ class AzDriver implements DriverModel {
     return `https://${this.accountName}.blob.core.windows.net/${this.bucket}/`
   }
 
-  listBlobs(prefix: string, page: ?string) {
+  listBlobs(prefix: string, page: ?string) : Promise<{ entries: Array<string>, page: any }> {
```
jcnelson: Can we create a `type` for `{ entries: Array<string>, page: any }` and use that here, instead of repeating this structure in all the drivers?  You can add it to the driver model.

hub/src/server/drivers/diskDriver.js
```diff
@@ -1,5 +1,5 @@
 /* @flow */
-import fs from 'fs'
+import fs from 'fs-extra'
```
jcnelson: The changes in this file are part of a different PR.  Please remove this commit from this PR.

hub/src/server/drivers/AzDriver.js
zone117x: Good idea 👍

admin/src/server.js
kantai: Agreed -- we should just rename the import.

hub/src/server/drivers/diskDriver.js
zone117x: There are a few minor changes in in this PR that require changes in the previous pending PR https://github.com/blockstack/gaia/pull/158 - easy to fix but maybe lets get that PR resolved first then I can update this one to accommodate?

admin/src/server.js
zone117x: Fixed

zone117x:
ping @kantai @jcnelson 
Would be nice to get this merged in for use in the feature branches

moxiegirl:
@zone117x Can you sign our contributor license agreement? Looks like we are near to merging this.

jcnelson:
Looks good to me.  Go ahead and merge :+1:

PR #172, opened 2019-01-02
Hotfix/y2k19 auth fix

jcnelson:
This patch enables authentication requests for 2018 to remain valid.

### DIFF
--- a/hub/package.json
+++ b/hub/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gaia-hub",
-  "version": "2.3.1",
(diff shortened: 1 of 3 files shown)


kantai:
Thanks for this patch @jcnelson -- I based a branch `hotfix/deprecate-timestamps` off this, that will remove the timestamp entirely going forward, which is probably a better approach.

kantai:
I am goign to merge this to master as this hotfix was already pushed to production. We can evaluate whether or not we want to merge the deprecation of timestamps separately.

PR #173, opened 2019-01-03
Deprecate timestamps

kantai:
This deprecates the annual timestamp in the challenge text -- it just replaces it with '0'.

This incorporates legacy years (2018, 2019) in challenge text checks as well.

### DIFF
--- a/hub/package-lock.json
+++ b/hub/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "gaia-hub",
-  "version": "2.2.2",
+  "version": "2.3.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


jcnelson:
Is my understanding correct in that a Gaia auth token is basically valid forever if it sets its span to `0`?  Or do Gaia auth tokens still expire (or have a revocation option)?

kantai:
There's an `exp` field in the JWT for setting an expiration time which Gaia auth respects. Legacy tokens _don't_ have such a field, however, legacy tokens are problematic in other ways, and haven't been generated in any libraries for at least 6 months.

jcnelson:
Okay, sounds good.  As long as tokens expire per the `exp` field, this PR looks good to me :+1:

kantai:
Thanks for the review @jcnelson -- tagging @zone117x for a review as well before merging.

kantai:
Great, thanks for the reviews! Merging.

PR #174, opened 2019-01-07
Add support for explicit blacklisting of tokens

kantai:
This allows us to explicitly blacklist tokens.

You would use this, e.g., if you knew specific tokens were compromised, and you wanted to invalidate some, but not all, tokens.

### DIFF
--- a/hub/src/server/authentication.js
+++ b/hub/src/server/authentication.js
@@ -359,7 +359,7 @@ export function parseAuthHeader(authHeader: string) {
   }
 }
 
-export function isBlackListedAuthorization(authHeader: string, blacklist: ?Map<string, boolean>) {
+export function isBlackListedAuthorization(authHeader: string, blacklist: ?Set<string>) {
   if (!blacklist)
     return
   if (!authHeader || !authHeader.toLowerCase().startsWith('bearer')) {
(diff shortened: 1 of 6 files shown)


hub/src/server/authentication.js
```diff
@@ -359,16 +359,31 @@ export function parseAuthHeader(authHeader: string) {
...
 }
 
+export function isBlackListedAuthorization(authHeader: string, blacklist: ?Map<string, boolean>) {
+  if (!blacklist)
+    return
+  if (!authHeader || !authHeader.toLowerCase().startsWith('bearer')) {
+    throw new ValidationError('Failed to parse authentication header.')
+  }
+  const authPart = authHeader.slice('bearer '.length)
+  if (blacklist.has(authPart)) {
```
hstove: Because we're strictly making the `blacklist` Map from an array, they'll always have a value of `true`. But `Map.has(key)` will return true, even if the blacklist has the value `key: false`. It's a little confusing here.

Maybe use a `Set`?

~~~javascript
blacklist = new Set()
blacklist.add('key')
blacklist.has('key)
~~~

kantai: Ah! I didn't know about `Set`s, that's exactly what I wanted.

kantai:
I think we don't quite want to implement this feature this way at the moment.

zone117x:
I think we'll need @jcnelson's review and approval here. I'm not familiar enough with the systems we are trying to fix with this.

PR #175, opened 2019-01-08
Hotfix for promise rejection behavior

kantai:
This wraps the server handler functions in promise.resolve so that exceptions thrown in the handler are caught by the `http` object's `.catch` clauses.

This has the effect of correctly returning errors messages that used to return 500s.

Added a test case to the `testHttp` unit tests to deal with this.

### DIFF
--- a/hub/CHANGELOG.md
+++ b/hub/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.3.3]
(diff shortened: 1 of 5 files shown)


zone117x:
Test of the fix looks good.

It's hard to reason about js exception behavior when Promises are used with callbacks. 
Especially this specific type of bug with sync exceptions combined with Promise chains, and having to manually instantiate Promise objects to fix it. 

I'm strongly in favor of using async/await syntax in the future. Much easier to reason about and avoid this type of js behavior.

kantai:
> I'm strongly in favor of using async/await syntax in the future. Much easier to reason about and avoid this type of js behavior.

Agreed -- async/await has been supported in node's LTS releases for a while now, so I see no real downside to it on server side code. When it required including the babel runtime stuff, I'm more against it.

PR #194, opened 2019-01-24
In-memory-only test driver & code cleanup

zone117x:
* An implementation of an in-memory driver (for testing not production) needed for good e2e testing of revocation.
* Added & cleaned up several Flow typings.
* Changed some promise-callback syntax to async/await syntax (readability and less error prone). 

Creating this before the main token revocation PR in an attempt to make reviewing that PR easier.


hub/test/src/testServer.js
```diff
@@ -230,11 +234,11 @@ export function testServer() {
 
     const authorization = `bearer ${authPart}`
     const authenticator = auth.parseAuthHeader(authorization)
-    t.throws(() => authenticator.isAuthenticationValid(testAddrs[1], challengeText),
+    t.throws(() => authenticator.isAuthenticationValid(testAddrs[1], [challengeText]),
```
zone117x: Improving the Flow typing in our tests showed a bug where this test was passing with an unintended error. (`String.includes` vs `Array.includes`)

hub/test/src/testDrivers/InMemoryDriver.js
```diff
@@ -0,0 +1,65 @@
+/* @flow */
```
jcnelson: Is this driver used anywhere else besides the unit test?  If the only point to having this driver is to test the Gaia hub, then why have it at all?  Why not use the disk driver for testing Gaia?  If this is only for testing, then why have the in-memory driver in the main source tree, instead of within the test itself?

hub/test/src/testDrivers.js
```diff
@@ -394,7 +390,41 @@ function testGcDriver() {
   })
 }
 
+function testInMemoryDriver() {
```
jcnelson: One gotcha we've had in the past is that people will store different kinds of data (with different `content-type`s), and expect to get the same data back.  This test only checks `application/octet-stream` for plain text.  If we keep the in-memory driver, can this test be expanded to include binary data and structured data?

hub/test/src/testDrivers.js
```diff
@@ -394,7 +390,41 @@ function testGcDriver() {
...
+          { path: 'foo.txt',
+            storageTopLevel: '12345',
+            stream: s,
+            contentType: 'application/octet-stream',
+            contentLength: contentBuff.length })
+
+      t.ok(readUrl.startsWith(prefix + '12345'), `${readUrl} must start with readUrlPrefix ${prefix}12345`)
+      const resp = await fetch(readUrl)
+      const resptxt = await resp.text()
+      t.equal(resptxt, 'hello world', `Must get back hello world: got back: ${resptxt}`)
```
jcnelson: This test also needs to verify that the content type matches what was given on the call to `performWrite`.

hub/src/server/utils.js
```diff
@@ -14,3 +17,46 @@ export function getDriverClass(driver: string) {
   }
 }
 
+
+export function readStream(stream: Readable, size: number): Promise<Buffer> {
```
jcnelson: Is this method used anywhere besides the in-memory driver?  If not, then it should be moved to that driver's file.

hub/test/src/testHttp.js
```diff
@@ -139,3 +227,9 @@ export function testHttp() {
       })
   })
 }
+
+export function testHttp() {
+  testHttpWithAzure()
+  testHttpDriverOption()
+  testHttpWithInMemoryDriver()
```
jcnelson: If we decide to drop the in-memory driver, then we should consider using a test with the disk driver (in a random direcory) instead.

hub/src/server/server.js
```diff
@@ -7,22 +7,28 @@ import { ProofChecker } from './ProofChecker'
...
-              config: { whitelist: Array<string>, serverName: string,
-                        readURL?: string, requireCorrectHubUrl?: boolean,
-                        validHubUrls?: Array<string> }) {
+
+  constructor(driver: DriverModel, proofChecker: ProofChecker, config: HubServerConfig) {
     this.driver = driver
     this.proofChecker = proofChecker
     this.whitelist = config.whitelist
-    this.serverName = config.serverName
+    this.serverName = (config.serverName: any)
```
kantai: why does serverName need to be `any` ?

zone117x: Existing test code is dependent on `serverName` coercing to `"undefined"`.
Perhaps we can update that test code to not depend on that instead?

hub/test/src/testHttp.js
zone117x: Agreed. These tests are for http not any specific driver, the existing tests just happened to be using the mock-azure config, so I named the function appropriately.

hub/src/server/server.js
kantai: Ah, yeah, let's do that

hub/src/server/utils.js
zone117x: That seems to be the case. Will move it. 👍

hub/test/src/testDrivers.js
zone117x: Good idea 👍

hub/test/src/testServer.js
kantai: Yep, good catch

hub/src/server/utils.js
kantai: I believe this is going to get used by the revocation PR.

hub/test/src/testDrivers/InMemoryDriver.js
zone117x: This started off as extensions to the existing `MockDriver` and eventually I moved it into its own class. There are a couple things this driver does only for the purpose of testing e2e components (like auth token revocation), such as providing its own http read-end point. The revocation touches both driver code and authentication, which means that it needs to have a decent mock driver that can cache a value. I think using the disk driver would involve injecting/wrapping much of this test code around it.

kantai: I think that the point is to be able to use it as a mock driver in unit tests --- using specific drivers for that is a pain, and conflates units tests with integration tests. The disk driver, for example, requires running a separate read endpoint, which you don't want to have to spin up to run unit tests. The drivers should be tested to ensure that they uphold the driver model separately --- the rest of the code should be able to use a mock driver (i.e., the in memory driver). Integration tests, however, shouldn't use the in memory driver.

> If this is only for testing, then why have the in-memory driver in the main source tree, instead of within the test itself?

It's in the test source directories, and could be consumed by multiple tests, it seems fine to pull it into its own test file.

jcnelson: Gotcha.  Thanks for the clarification :+1:

jcnelson:
Looks good to me :+1: Thanks @zone117x for all your work on this PR!

Issue #150, opened 2018-10-23
Support for auth token revocation


jehunter5811:
Some apps may build OAuth style API solutions. Currently, in doing so, access to the authToken means that whatever outside app is using the blockstack app's API will have access to that token for 4 months. The end user should have a way to either set the expiry timeframe or revoke the token. 

This is outside the association token concept. This is for the core authToken.

hstove:
> The end user should have a way to either set the expiry timeframe

The [current authentication scheme](https://github.com/blockstack/gaia#v1-authentication-scheme) supports passing an `exp` into the signed JWT used. You can use this to make it expire.

Supporting revocation of an existing key would require some new features, and would require a Gaia hub to keep some state about current and revoked keys, which it does not do.

hstove:
I would like to see revocation of tokens, though. Kanstack will have collaboration soon, and it uses scoped auth tokens to write to a shared bucket. However, there is still an issue when you want to 'remove' a shared collaborator. I have mechanisms to rotate a shared private key, but the old gaia token is still valid. You could rotate the scoped auth token, I suppose.

One way of managing state could be to store revoked tokens in the user's hub. This would require Gaia to read from some predetermined location in the hub in order to see if a token is revoked. That could work, but it'd make writes slower by adding a network request. I'm not sure that's the best solution. Requiring hubs to use a database to store state wouldn't work, because migrating to a different hub would require migrating the database as well.

In fact, I'm thinking of dropping the usage of a shared group hub, and instead using signatures and an indexer to do collaboration. That solves the issue of not having the ability to revoke tokens.

kantai:
Something we can do (with the help of a stateful sidecar) is support something like a "bump" to revoke, where we keep a single nonce in the stateful sidecar, which all tokens must use. Then, to revoke tokens, the stateful sidecar could bump that nonce. This would revoke _all outstanding tokens_, but would probably be the simplest to reason about. (Revoking individual tokens isn't hard to implement on the server, but a UI would need to assist users in figuring out which auth tokens even exist).

kantai:
### Supporting revocation

Roughly speaking, we want _something_ like the following properties for revocation.

1. Revocation on at least a _per-bucket_ basis. Something like "revoke all tokens for the entire gaia hub" doesn't work because of shared gaia hubs.
2. Revocation must be performable client-side, with the gaia bucket's private key.
3. (Probably) Revocation should not require too much server state. That means, this would probably need to be stored _using the driver itself_.

[...]

PR #196, opened 2019-01-28
Token revocation feature

zone117x:
Requesting feedback on:
* Are the correct public addresses being checked / validated for?
* Is the token timestamp creation and validation secure?

https://github.com/blockstack/gaia/issues/150


hub/src/server/authentication.js
```diff
@@ -81,9 +83,15 @@ export class V1Authentication {
...
 
+    let payloadCreationDate = (Date.now()/1000|0)
+    if (creationDate){
+      payloadCreationDate = creationDate
+    }
+
     const payload: TokenPayloadType = { gaiaChallenge: challengeText,
                       iss: publicKeyHex,
                       exp: FOUR_MONTH_SECONDS + (new Date()/1000),
+                      creationDate: payloadCreationDate,
```
jcnelson: Can we use standard JWT fields like `iat` for "issued at" and `nbf` for "not used before" to encode this? 
 Full list here:  https://en.wikipedia.org/wiki/JSON_Web_Token

hub/src/server/authentication.js
```diff
@@ -230,6 +240,23 @@ export class V1Authentication {
       throw new ValidationError('Must provide `iss` claim in JWT.')
     }
 
+    // check for revocations
+    if (options && options.oldestValidTokenTimestamp && options.oldestValidTokenTimestamp > 0) {
```
jcnelson: We're probably going to have to schedule a "flag day" whereby all future tokens need to have issued-at timestamps.  Maybe that can be encoded here?  i.e. we allow you to use a token without a timestamp up until a given date, and then require it?  This way, we can roll out the Gaia hub to everyone before rolling out the required changes to blockstack.js.

jcnelson: Or maybe I'm misunderstanding -- is `oldestValidTokenTimestamp` taken from the Gaia hub's config file?

hub/src/server/http.js
```diff
@@ -107,6 +109,45 @@ export function makeHttpServer(config: MakeHttpServerConfig & HubServerConfig) :
...
+        return
+      }
+
+      if (!req.body || !req.body.oldestValidTimestamp) {
+        writeResponse(res, { mesasge: 'Invalid JSON: missing oldestValidTimestamp'}, 400)
+        return
+      }
+
+      const address = req.params[0]
+      const oldestValidTimestamp: number = parseInt(req.body.oldestValidTimestamp)
```
jcnelson: This needs to be surrounded with try/catch in case the caller doesn't send a number here.

hub/src/server/authentication.js
zone117x: `options.oldestValidTokenTimestamp` is coming from the special auth file stored in the gaia bucket / LRU cache. 
So currently, if bucket owner has not issued any revocations, then the token issued-date will not be checked.

hub/src/server/authentication.js
```diff
@@ -81,9 +83,15 @@ export class V1Authentication {
       validateScopes(scopes)
     }
 
+    let payloadCreationDate = (Date.now()/1000|0)
```
hstove: Nitpick: could just be `const paloadCreationDate = creationDate || (Date.now()/1000|0)`

zone117x:
Thanks @hstove - I added documentation for the new http endpoint and JWT field. 
And yeah I agree that blockstack.js should just POST Date.now() to the `revoke-all` endpoint.

hub/src/server/authentication.js
jcnelson: Yup, that sounds even better :+1:

hub/src/server/http.js
jcnelson: Lol, once again I set my expectations too high for JS.  Thanks for clarifying :+1:

hub/src/server/revocations.js
```diff
@@ -0,0 +1,98 @@
...
+const AUTH_TIMESTAMP_FILE_NAME = 'authTimestamp'
+
+export class AuthTimestampCache {
+
+  cache: LRUCache;
+  driver: DriverModel;
+
+  constructor(driver: DriverModel) {
+    // TODO: configure cache settings..
+    this.cache = new LRUCache()
```
jcnelson: How big can this cache get?  Can it exceed the size of RAM?

jcnelson: Also, just want to confirm -- the `LRUCache` is _only_ in-RAM, right?

zone117x: Reasonable cache max default has been implemented, with eviction warning logging, and tests to ensure cache eviction is working as expected.

hub/src/server/revocations.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  async readAuthTimestamp(bucketAddress: string): Promise<number> {
+
+    const readUrlPrefix = this.driver.getReadURLPrefix()
+    const authTimestampDir = this.getAuthTimestampFileDir(bucketAddress)
+    
+    // Check if an auth number file exists
+    // TODO: Is the error for "file not exists" consistent enough to depend upon instead of doing this?
+    const bucketFileList = await this.driver.listFiles(authTimestampDir, null)
```
kantai: Why do we do a listFiles here? Why not just let the subsequent fetch throw an error, and return `0` ?

hub/src/server/revocations.js
```diff
@@ -0,0 +1,121 @@
...
+
+    try {
+      const authNumberFileUrl = `${readUrlPrefix}${authTimestampDir}/${AUTH_TIMESTAMP_FILE_NAME}`
+      const fetchResult = await fetch(authNumberFileUrl)
+      const authNumberText = await fetchResult.text()
+      const authNumber = parseInt(authNumberText)
+      return authNumber
+    } catch (err) {
+      // TODO: If error indicates file does not exit then return 0 (indicates no auth number setup).
+      throw new Error('Unimplemented')
```
kantai: This should `return 0`, right?

hub/src/server/revocations.js
jcnelson: Yeah, this shouldn't use `listFiles` if at all possible.  `listFiles` isn't guaranteed to be consistent or performant across all the storage systems we'd like to use.

hub/src/server/revocations.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  getAuthTimestampFileDir(bucketAddress: string) {
+    return `${bucketAddress}-auth`
+  }
+
+  async readAuthTimestamp(bucketAddress: string): Promise<number> {
+
+    const readUrlPrefix = this.driver.getReadURLPrefix()
```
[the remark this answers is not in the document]
hstove: On my Digital Ocean hub, I needed it to override and use `server.getReadUrlPrefix()` because I'm using the s3 driver but not with S3.

hub/src/server/revocations.js
zone117x: That's what I suspected. I'm performing the fetch below here, but wasn't confident about what to do in any given error context. What happens if it is just a http timeout but there is a revocation timestamp? Then returning 0 would allow any potentially revoked tokens to work.

hub/src/server/revocations.js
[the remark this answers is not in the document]
zone117x: @hstove so are you in agreement with @kantai that this line should use `server.getReadUrlPrefix()` instead?

hub/src/server/revocations.js
jcnelson: If there's an HTTP timeout in loading the revocation timestamp, and the token has revocation data (i.e. an `iat` field), then I think the entire request should fail with 403 -- the caller is clearly asking for the revocation logic to be applied, and the authentication code's inability to be certain about the validity of the token under these circumstances implies that the _safest_ default behavior should be taken.

kantai: I think a `404` should result in a `0` here.

In all likelihood, clients will _always_ set an `iat` field when generating tokens in the future. In the case that they haven't called `/revoke-all/` yet, all of their calls will 403 (even calls to `revoke-all`!) I think that's clearly bad behavior. A `404` should only occur if the file really isn't there -- in the event of a malicious storage provider, none of this matters anyways, they can write data whenever they want regardless of token validity.

hub/src/server/revocations.js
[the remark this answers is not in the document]
zone117x: This has been updated to use `server.getReadUrlPrefix()`

hub/src/server/revocations.js
zone117x: The logic around here has been flushed out.

hub/src/server/revocations.js
zone117x: This has been updated to no longer user `driver.listFiles()` and instead uses the fetch response - in accordance to what we discussed.

hub/test/src/testServer.js
```diff
@@ -212,6 +212,48 @@ export function testServer() {
...
+    authPart = auth.V1Authentication.makeAuthPart(testPairs[0], challengeText, undefined, undefined, undefined, futureDate + 10000)
+    authorization = `bearer ${authPart}`
+  
+    // request should succeed with a token iat newer than the revocation date
+    await server.handleRequest(testAddrs[0], '/foo/bar', 
+                              { 'content-type' : 'text/text',
+                                'content-length': 400,
+                                authorization }, getJunkData())
+  })
+
```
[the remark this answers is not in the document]
zone117x: These tests have been added - please review that they perform the intended checks

hub/test/src/testServer.js
```diff
@@ -10,373 +10,545 @@ import { DriverModel } from '../../src/server/driverModel'
...
 
-    const writeScopes = [
-      {
-        scope: 'putFile',
-        domain: '/foo/bar',
-      },
-      {
-        scope: 'putFilePrefix',
-        domain: 'baz'
+  test('association tokens with a whitelisted signer can bump any address', async (t) => {
```
kantai: I don't understand this test -- it seems like the following is happening:

1. An association token is created and included in an authorization token -- this conveys roughly the following:
    Token grants access to write to testPairs[0], testPairs[0] is not in the whitelist, but testPairs[1] is, and grants testPairs[0] the ability to use the gaia hub.
2. That authorization token is used to bump the revocation number for testPairs[0]

Right? In that case, it doesn't really match 'association tokens with a whitelisted signer can bump any address', which isn't behavior we want anyways, right?

jcnelson: > Right? In that case, it doesn't really match 'association tokens with a whitelisted signer can bump any address', which isn't behavior we want anyways, right?

No, I think it is.  You need the ability to invalidate auth tokens on a per-app basis.  This can only be achieved on a private Gaia hub (i.e. with a non-empty whitelist) if one of two things are true:

* The app address is whitelisted
* The owner address is whitelisted, and the auth token has an association token authorizing the bump on the app address

This requirement translates to "an auth token with an association token signed by a whitelisted address can be used to bump the auth token's signer's address" -- which is to say, a whitelisted address can bump _any_ address's auth revocation timestamp on the Gaia hub.

jcnelson: However, this is _not_ to say that an auth token can be used to bump any address -- regardless of whether or not it has an association token, an auth token can only bump the address of the key that signed it.

kantai: > This requirement translates to "an auth token with an association token signed by a whitelisted address can be used to bump the auth token's signer's address" -- which is to say, a whitelisted address can bump any address's auth revocation timestamp on the Gaia hub.

Hmm, it still should only ever be the case that a signer can only bump the revocation timestamp for its corresponding address -- and it appears this is what is implemented in the PR. The call to `validate` will only pass if the auth JWT is signed by the private key corresponding to the address being validated.

It's true that if the whitelisted address is a parent key of another address, it'd be able to generate such a token, but that's outside the purview of the authentication scheme in Gaia. But this isn't the case for all addresses, anyways --- imagine there are two whitelisted "owners" on the same private hub -- one owner wouldn't be able to generate auth tokens to bump revocation numbers for app-keys of the other owner.

> However, this is not to say that an auth token can be used to bump any address -- regardless of whether or not it has an association token, an auth token can only bump the address of the key that signed it.

Right, I think that's the same as what I'm saying above.

jcnelson: Looking at the test, it's only confirming that a request to bump the non-whitelisted address will be honored as long as the auth token has a well-formed association token signed by a whitelisted address.

kantai: Yep, I think naming it  "Non-whitelisted address can bump revocation if bearing valid association token" would match what its testing

zone117x: Test renamed

jcnelson:
I'm looking through the code to do a final pass.  Looks good!  However, there are a few things that should be fixed:

* `handleListFiles()` should honor the revocation timestamp, but it does not do so right now.
* The default configuration in `config.js` needs a value for `authTimestampCacheSize`
* `writeAuthTimestamp()` should call `this.cache.set()` _after_ both verifying that the data is small enough and well-formed, and ensuring that it was successfully stored to the backend data store (with `driver.performWrite()`).
* `writeAuthTimestamp()` should return a `ValidationError` if the new timestamp is too big.  Right now it returns a generic `Error`, which triggers a 500 response (whereas it should be a 401).

zone117x:
@jcnelson all issues pointed point should be addressed now.

> The default configuration in config.js needs a value for authTimestampCacheSize

Note: this was already in there, is there a different file or area where the default is missing?

hub/src/server/revocations.js
```diff
@@ -0,0 +1,147 @@
...
+    this.cache.set(bucketAddress, authTimestamp)
+
+    return authTimestamp
+  }
+
+  async writeAuthTimestamp(bucketAddress: string, timestamp: number) : Promise<void> {
+
+    // Recheck cache for a larger timestamp to avoid race conditions from slow storage.
+    const cachedTimestamp = this.cache.get(bucketAddress)
+    if (cachedTimestamp && cachedTimestamp > timestamp) {
```
jcnelson: The check here, as well as the `this.cache.set()` call, need to happen _after_ the call to `this.driver.performWrite()`.

hstove:
Random thought, and we shouldn't go down this road right now. Since we're implementing a cache, it'd be good to have other cache drivers, like Redis. Then, the cache wouldn't reset when different the server restarts, and reduce memory usage, and allow us to use our new Redis cluster.

zone117x:
@hstove redis is a great idea. We can pretty easily make the cache an interface that defaults to the current implementation.

hub/src/server/revocations.js
zone117x: This should be fixed now.

jcnelson:
Looks good to me :+1: @hstove can @zone117x merge this?

zone117x:
Going to run the driver tests against the real cloud APIs to ensure the http status codes behave as expected then merge in.

Issue #202, opened 2019-02-06
Need a JSON schema for our configuration parameters


moxiegirl:
I manually created one but I doubt it is complete.  @kantai 

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "servername": { "type": "string" },
    "port": { "type": "integer" },
    "driver": { "type": "string" },
    "bucket": { "type": "string" },
    "readURL": { "type": "string" },
    "requireCorrectHubUrl": { "type": "string" },
    "awsCredentials": {
      "type": "object",
      "properties": {
        "endpoint": { "type": "string"},
        "accessKeyId": { "type": "string"},
        "secretAccessKey": { "type": "string"}
      },
      "required": [
        "endpoint",
        "accessKeyId",
        "secretAccessKey"
      ]
    },
    "proofsConfig": {
      "type": "object",
      "properties": {
        "proofsRequired": { "type": "string"}
      },
      "required": [
        "proofsRequired"
      ]
    },
    "azCredentials": {
      "type": "object",
      "properties": {
        "accountName": { "type": "string"},
        "accountKey": { "type": "string"}
      },
      "required": [
        "accountName",
        "accountKey"
      ]
    },
    "gcCredentials": {
      "type": "object",
      "properties": {
        "keyFilename": { "type": "string"}
      },
      "required": [
        "keyFilename"
      ]
    },
    "diskSettings": {
      "type": "object",
      "properties": {
        "storageRootDirectory": { "type": "string"}
      },
      "required": [
        "storageRootDirectory"
      ]
    },
    "pageSize": { "type": "integer"},
    "argsTransport": {
      "type": "object",
      "properties": {
        "level": { "type": "string"},
        "handleExceptions": { "type": "boolean" },
        "stringify": { "type": "boolean" },
        "timestamp": { "type": "boolean" },
        "colorize": { "type": "boolean" },
        "json":{ "type": "boolean" }
      },
      "required": [
        "level",
        "handleExceptions",
        "stringify",
        "timestamp",
        "colorize",
        "json"
      ]
    }
  },
  "required": [
    "servername",
    "port",
    "driver",
    "bucket",
    "readURL",
    "requireCorrectHubUrl",
    "awsCredentials",
    "proofsConfig",
    "azCredentials",
    "gcCredentials",
    "diskSettings",
    "pageSize",
    "argsTransport"
  ]
}
```

moxiegirl:
[...]

PR #204, opened 2019-02-12
Config schema enforcement

kantai:
This add (pretty light) JSON schema enforcement to the hub's config JSON.

Any currently working configuration should continue to work -- though if the config file contained invalid or unused properties, that file would no longer work.

We don't use `required` parameters for the most part in this file, because any of these settings could instead be set via environment variables.

I added test cases to cover sample configs and test enforcment of a bad JSON file.

### DIFF
--- a/hub/src/server/config-schema.json
+++ b/hub/src/server/config-schema.json
@@ -29,7 +29,7 @@
       {
         "type": "object",
         "properties": {
-          "proofsRequired": { "type": "string"}
+          "proofsRequired": { "type": "integer"}
         },
         "required": [
           "proofsRequired"
--- a/hub/src/server/config-schema.json
+++ b/hub/src/server/config-schema.json
@@ -3,12 +3,12 @@
   "type": "object",
   "properties": {
     "serverName": { "type": "string" },
-    "port": { "type": "integer" },
+    "port": { "type": "integer", "minimum": 0, "maximum": 65535 },
     "driver": { "type": { "enum":
                           [ "aws", "azure", "disk", "google-cloud" ] } },
     "bucket": { "type": "string" },
     "readURL": { "type": "string" },
-    "pageSize": { "type": "integer"},
+    "pageSize": { "type": "integer", "minimum": 1, "maximum": 4096 },
     "requireCorrectHubUrl": { "type": "boolean" },
     "cacheControl": { "type": "string" },
     "whitelist": { "oneOf": [
--- a/hub/src/server/config-schema.json
+++ b/hub/src/server/config-schema.json
@@ -1,0 +1,88 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "serverName": { "type": "string" },
+    "port": { "type": "integer" },
+    "driver": { "type": { "enum":
+                          [ "aws", "azure", "disk", "google-cloud" ] } },
+    "bucket": { "type": "string" },
+    "readURL": { "type": "string" },
+    "pageSize": { "type": "integer"},
+    "requireCorrectHubUrl": { "type": "boolean" },
+    "cacheControl": { "type": "string" },
+    "whitelist": { "oneOf": [
+      {
+        "type": "array",
+        "items": { "type": "string" }
+      },
+      { "type": ["boolean", "null"] }
+    ]},
+    "validHubUrls": { "oneOf": [
+      {
+        "type": "array",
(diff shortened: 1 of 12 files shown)


kantai:
This addresses #202

hub/src/server/config-schema.json
```diff
@@ -0,0 +1,88 @@
...
+        "type": "array",
+        "items": { "type": "string" }
+      },
+      { "type": ["boolean", "null"] }
+    ]},
+    "proofsConfig": { "oneOf": [
+      {
+        "type": "object",
+        "properties": {
+          "proofsRequired": { "type": "string"}
```
jcnelson: Shouldn't this be an integer?

hub/src/server/config-schema.json
```diff
@@ -0,0 +1,88 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "serverName": { "type": "string" },
+    "port": { "type": "integer" },
+    "driver": { "type": { "enum":
+                          [ "aws", "azure", "disk", "google-cloud" ] } },
```
jcnelson: Is there a way this list of drivers can be DRY'ed up?  There's other code (like the server initialization code) that selects which module to load based on a driver name.

hub/src/server/config-schema.json
```diff
@@ -0,0 +1,88 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "serverName": { "type": "string" },
+    "port": { "type": "integer" },
```
jcnelson: Should probably put min and max values here.

hub/src/server/config-schema.json
```diff
@@ -0,0 +1,88 @@
...
+    "driver": { "type": { "enum":
+                          [ "aws", "azure", "disk", "google-cloud" ] } },
+    "bucket": { "type": "string" },
+    "readURL": { "type": "string" },
+    "pageSize": { "type": "integer"},
```
jcnelson: This should be a positive integer, and probably not too big (i.e. no bigger than 4096).

hub/src/server/config-schema.json
```diff
@@ -0,0 +1,88 @@
...
+      }
+    },
+    "azCredentials": {
+      "type": "object",
+      "properties": {
+        "accountName": { "type": "string"},
+        "accountKey": { "type": "string"}
+      }
+    },
+    "gcCredentials": {
```
jcnelson: For each of these driver schemas (not just `gcCredentials`), do we know which fields are required?

hub/src/server/config-schema.json
kantai: DRY'ed ?

hub/src/server/config-schema.json
kantai: Yep

hub/src/server/config-schema.json
kantai: None of them are required. You can set almost all of them via provider specific environment variables (i.e., via the aws, azure, or gcp CLIs) or via Gaia environment variables.

hub/src/server/config-schema.json
jcnelson: DRY == Don't Repeat Yourself.

jcnelson: On that note, we might want to consider moving the driver-specific aspects of this schema to the driver model.

kantai: Well, most of the point of providing this file is to have a schema file that we can link to or explicitly include in documentation -- for that, having a whole schema file is important.

We _could_ make this file auto-generate on build and commit the result.

jcnelson: Ah, I see.  I'm fine with this for now as long as the number of drivers is small, but once we exceed ~10 of them we might want to reconsider auto-generating the schema.

hub/src/server/config-schema.json
jcnelson: Gotcha :+1:

moxiegirl:
For the documentation release, I took a stab at the config page.  I need to expand it with defs of each section

https://docs.blockstack.org/storage/config-schema.html

There is a how to section also 

https://docs.blockstack.org/storage/gaia-admin.html

zone117x:
Close this in favor of https://github.com/blockstack/gaia/pull/229 ?

PR #206, opened 2019-02-12
Feature/docker compose

wileyj:
Most changes are for new `docker` dir ./docker. 
- fixed a few code typos in the admin/src/server.js file relating to json config schema
- added some dockerfiles for building the different containers (admin, reader, hub)
- all other files are related to running a gaia hub with docker (nginx, certbot, hub, reader, admin images)


docker/nginx/conf.d/nginx-ssl.conf
```diff
@@ -0,0 +1,101 @@
...
+  #security headers
+  add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload";
+  add_header X-XSS-Protection "1; mode=block" always;
+  add_header X-Content-Type-Options "nosniff" always;
+  add_header X-Frame-Options "DENY" always;
+
+  ##CSP
+  add_header Content-Security-Policy "frame-src 'self'; default-src 'self'; script-src 'self' 'unsafe-inline' https://maxcdn.bootstrapcdn.com https://ajax.googleapis.com; img-src 'self'; style-src 'self' https://maxcdn.bootstrapcdn.com; font-src 'self' data: https://maxcdn.bootstrapcdn.com; form-action 'self'; upgrade-insecure-requests;" always;
+  add_header Referrer-Policy "strict-origin-when-cross-origin" always;
+
```
kantai: Okay -- I think there's a couple of issues here -- 

The following paths need to go the `gaia_hub` upstream:

`/store/*`, `/hub_info`, `/list-files/*` (and, in the future `/revoke/`)

Alternatively, we can say, send all `/hub/*` traffic to the `gaia_hub` upstream, then delete the `/hub_info` entry from here.

That configuration would give us:

`/hub/` -> points to the gaia hub, which services writes

`/reader/` -> points to the reader. which services reads

docker/hub-config/config.json
```diff
@@ -0,0 +1,29 @@
+{
+  "serverName": "hub",
+  "port": 3000,
+  "driver": "disk",
+  "bucket": "hub",
+  "readURL": "http://localhost:80/hub/",
```
kantai: This needs to be set to the DNS hostname, the reader prefix, and should be a `https` URL -- so, if I set up gaia.aaron.com to at my VM -- this would need to be:

```
"https://gaia.aaron.com/reader/"
```

docker/nginx/conf.d/nginx-ssl.conf
wileyj: I"ll work on the nginx config here. i think i understand what you're suggesting and will update later.

docker/hub-config/config.json
wileyj: this will require some changes to the image ignition script. 
it was planned already for the admin API_KEY, so i'll do that and the readURL based on this feedback.

hub/README.md
```diff
@@ -1,3 +1,5 @@
+The rest of this will probably be refactored
+
 # Running the Gaia Hub
```
moxiegirl: This README actually says further down:

```
The Gaia hub currently supports the following drivers:

```
'aws' == Amazon S3
'azure' == Azure Blob Storage
'disk' == Local disk (you must set up static web-hosting to point at this driver)
'google-cloud' === Google Cloud Storage
```

Set the driver you wish to use in your `config.json` file with the `driver` parameter. Many drivers additionally accept the `bucket` parameter, which controls the bucket name that files should be written to.
```

Do we actually want to publish this in our README this way without a caveat to testing and so forth?

wileyj:
merged testing branch into this PR to address comments

docker/docker-compose.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+  admin:
+    image: ${ADMIN_IMAGE}
+    restart: always
+    env_file:
+      - .env
+    volumes:
+      - "./admin-config:/src/admin/etc"
+      - "./hub-config:/tmp/hub-config"
+      - "/var/run/docker.sock:/var/run/docker.sock"
```
kantai: Such a cool way to access docker from inside the container

docker/hub-config/config.json
wileyj: addressed in: https://github.com/blockstack/gaia/blob/feature/docker-compose/docker/update_configs.sh

which is called in the gaia-hub service

hub/README.md
wileyj: that line was added here:
https://github.com/blockstack/gaia/commit/f912de9364aac4cce67722718e9656d98bfe8565#diff-f544082c9f728f2d36ca08fe1a5843f3

@GInxh , not sure of the reasoning behind this addition.

docker/AWS.md
```diff
@@ -0,0 +1,133 @@
...
+Mainly, it's going to be pulling down all the docker images specified in the docker-compose files in this repo, and can take a little bit of time to complete.
+
+We do this to ensure that when you start the VM, you're getting the most up to date versions of the image at the time you start the instance.
+
+---
+Final Steps:
+
+Command to confirm your DNS is set correctly:
+    - this will check every 5 seconds for a change in output:
+        - `watch -n 2 -t -g -x host <domain>
```
zone117x: Minor markdown formatting issues around here

zone117x:
Just noticed this is going to master. Shouldn't it be to the dev branch?

kantai:
Things like this should appropriately hit `master`, I think. It's infrastructure setup code (which is in production on the published AMIs.)

Issue #197, opened 2019-02-01
Test drivers against live cloud APIs on CircleCI


zone117x:
We need test API keys & buckets generated for azure, google cloud, and S3.
These will be encrypted and added to the circleci config file.

zone117x:
Pinging @wileyj - if you can send me the cloud API creds that are easy to get right now, then I can implement the tests for just the those. We can create separate issues for ones that will take more more work.

PR #210, opened 2019-02-26
Feature/cloud api testing

zone117x:
## Goal of PR

Perform integration testing for all drivers https://github.com/blockstack/gaia/issues/197. This PR gives us the following: 
* Assurance that the revocation methodology of depending on cloud API 404s works reliably. 
* Implement more tests for ensuring driver/cloud API compliance and behavioral consistency. 
* Easier to implement/update/test drivers: a pattern for specifying mock driver code and integration config setup code now exists. These are automatically ran against our test suite ensuring driver compliance. 
* A json object for all cloud API credentials are stored in a private CircleCI env var. 

## Implementation

Normalized the method for developers to add to the set of mock driver implementations and the set of driver integration setup methods.

Each driver implementation has an entry for both A) mock instance setup method, and B) an method that instantiates an production driver instance when the credentials/config are specified via env vars.  

Both these sets are iterated over, with each driver being instantiated then ran through the main driver unit test. In additional, several tests dealing with authentication and revocations are now also ran against all configured drivers. 
 
## Manual Testing

1. Git pull
1. Specify cloud API credentials using either the `DRIVER_CONFIG_TEST_DATA` env var pointed to the `Driver Credentials JSON` found in the `Gaia Driver Testing Keys` 1password vault (ask @kantai or myself for access). The circleCI tests are with this. 
   * Alternatively, the previous (driver specific) env vars can still be used. For the the list of these env vars see:  https://github.com/blockstack/gaia/blob/65ba7a5d945db1be458b78465bd9ce1a4f6ce223/hub/test/src/testDrivers/integrationTestDrivers.js#L24
    * If no cloud credentials are specified, the integration tests will still be ran using the self-hosted diskDriver and the InMemoryDriver. 
1. `npm i`
1. `npm test`

## Automated Testing

The disk driver now has a mock and self-hosted integration implemented. This, along with the InMemoryDriver, are always ran (no environment/external configuration needed). Addressing @jcnelson's earlier revocation PR comment about using the diskDriver rather than the InMemoryDriver - we can probably remove the InMemoryDriver now that we have this setup. However, this did create the first inter-project npm dev-dependency where the `hub` project now depends the the `reader` project. We should evaluate and confirm that we are OK with this.


zone117x:
Several bugs were identified and fixed while writing these tests. Some of the fixes involved changes that I would have preferred to have submitted separate PRs for, however, the many of them were tightly coupled with the changes required for the integration tests. The code coverage reports and significant increase in tests should hopefully make this PR review a bit easier. 

## Bug Fixes:

* Fix pagination/pageSize bugs with AzDriver, S3Driver, and diskDriver. 
* Fixed bug Azure and GoogleCloud drivers where `performWrite` did not propagate http upload stream errors - leaving hanging promises, event listeners, and/or open streams. 
* The disk driver contained a bug `performWrite` where the asynchronous pipe between the http POST stream and file-write-stream was created, abandoned, and a 2xx success http response being immediately sent. Any errors occurring during http upload or file writing would ignored. Also, this caused integration test failures when writing then reading since the file stream would not yet be flushed to disk.
* The disk driver also contained several synchronous `fs` usages which could freeze up the main event loop under certain types of disk load. A mounted network drive comes to mind. 
* The `reader` server wasn't checking for path traversals (`..`) in the filename request. 
* Fixed a few edge-case errors occurring next to manual Promise chaining which would not get propagated through the async call stack.

reader/.flowconfig
```diff
@@ -8,5 +8,6 @@
 [libs]
 
 [options]
+server.max_workers=1
```
kantai: Why set the max_workers to 1? Wouldn't allowing more workers speed up the flow checks?

hub/test/src/testDrivers/integrationTestDrivers.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+export const driverConfigs = {
+  az: undefined,
+  aws: undefined,
+  gc: undefined,
+  disk: undefined
+};
+
+if (driverConfigTestData) {
```
kantai: Should we just use the env vars defined by the drivers/config.json itself to pass in the config data?

hub/test/src/testDrivers/integrationTestDrivers.js
```diff
@@ -0,0 +1,131 @@
...
+
+    // Pull in the gaia-reader http server.
+    //$FlowFixMe
+    const gaiaReader = require('gaia-reader/lib/http');
+    const app: express.Application = gaiaReader.makeHttpServer(selfHostedConfig);
+    const serverPromise = new Promise((res, rej) => {
+      const server = app.listen(0, 'localhost', err => err ? rej(err) : res(server));
+    });
+
+    return new class SelfHostedDiskDriver extends DiskDriver {
```
kantai: nice. this is a cool way to initialize the disk driver.

reader/.flowconfig
zone117x: Flow appears to perform "parallelism" by spawning new parsing threads on an interval. CircleCI broke on the flow parsing step after x amount of these workers were spawned and exhausted the memory. After trial and error this was the only thing that fixed it.

hub/test/src/testDrivers/integrationTestDrivers.js
zone117x: I mostly set it up this way to avoid having to setup lots of env vars locally or in a given CI system in order to perform all integration tests

kantai: Makes sense

hub/src/server/drivers/S3Driver.js
```diff
@@ -62,31 +71,57 @@ class S3Driver implements DriverModel {
     return `https://${this.bucket}.s3.amazonaws.com/`
   }
 
-  createIfNeeded () {
-    this.s3.headBucket( { Bucket: this.bucket }, (error) => {
-      if (error && error.code === 'NotFound') { // try to create
+  async createIfNeeded () {
+    try {
+      await this.s3.headBucket({ Bucket: this.bucket }).promise()
```
kantai: should this use `this.s3BucketExists()`  instead?

jcnelson: I'm also curious why `this.s3BucketExists()` isn't being used.

jcnelson:
Looks great to me! :tada: :tada: :tada:

hub/src/server/drivers/S3Driver.js
zone117x: Yep, fixed

moxiegirl:
@zone117x Do we want developers contributing to Blockstack to run these tests locally prior to submitting a PR --- and will they run automatically on a PR submission?  Is there a use case outside of Gaia contributors for running them?

zone117x:
No. Requiring contributors to procure accounts and credentials for all cloud services is not reasonable. The full integration tests will be ran automatically once they submit a PR. 

The default local tests (`npm test`) include integration tests using the disk driver and in-memory driver which should be good enough before submitting a PR.

PR #213, opened 2019-03-11
Typescript port

zone117x:
## Goal of PR

Port Flow js to typescript

## Implementation

For all three sub-projects (`hub`, `reader`, `admin`):
- [x] `src` compiling as typescript with `--noImplicitAny=true`
- [x] unit test source compiling as typescript
- [x] unit tests passing
- [x] update linting - port eslint js rules to equivalent ts rules
- [x] code coverage & codecov.io reporting

Performed a rebase of these commits on top of an initial commit that only performed `*.js` to `*.ts` file renames in order to preserve git history. Github diff may show src file deletes & additions but proper git history is good to go. 

## Manual Testing

`npm run test`

## Automated Testing

Coverage reports are now being generated for all subprojects. This is why total codecov reports lower total percent coverage. 

## Submission Checklist

- [x] The code passes our eslint definitions, unit tests, and contains correct type annotations.

- [x] Submission documents any new features or endpoints, and describes how developers would be expected to interact with them.


admin/src/server.ts
```diff
@@ -1,69 +1,70 @@
-/* @flow */
-
 import fs from 'fs'
-import logger from 'winston'
 import childProcess from 'child_process'
 import Ajv from 'ajv'
+import { Config, logger } from './config'
+
+function runSubprocess(
```
jcnelson: This should return one of the following:

```
{ statusCode: number, status: { result: string } }
```
or
```
{ statusCode: number, error: string }
```

Not sure if there's a way to capture this in Typescript?  Also, I'm happy to change the return type to something simpler if it would get us better type annotations.

hub/src/server/authentication.ts
```diff
@@ -349,19 +354,19 @@ export class LegacyAuthentication {
...
   isAuthenticationValid(address: string, challengeTexts: Array<string>,
-                        options? : {}) { //  eslint-disable-line no-unused-vars
+                        options? : {}) { //  eslint-disable-line @typescript-eslint/no-unused-vars
     if (ecPairToAddress(this.publickey) !== address) {
       throw new ValidationError('Address not allowed to write on this path')
     }
 
     for (const challengeText of challengeTexts) {
-      const digest = bitcoin.crypto.sha256(challengeText)
+      const digest = bitcoin.crypto.sha256(Buffer.from(challengeText))
```
jcnelson: Interesting.  Does this work the same as before, without `Buffer.from`?

hub/src/server/config.ts
```diff
@@ -54,7 +78,7 @@ function getConfigEnv(envVars) {
           throw new Error(`Passed a non-number input to: ${envVar}`)
         }
       } else if (parseLists.indexOf(name) >= 0) {
-        configEnv[name] = configEnv[name].split(',').map(x => x.trim())
+        configEnv[name] = (<string>configEnv[name]).split(',').map(x => x.trim())
```
jcnelson: I'm unfamiliar with this syntax.  Is this a typecast?

zone117x: Yep!

hub/src/server/authentication.ts
zone117x: The type definitions only accept a Buffer: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bitcoinjs-lib/index.d.ts#L342

Looking through the source it looks like it ends up using node's hash.update API https://nodejs.org/api/crypto.html#crypto_hash_update_data_inputencoding which does accept a string.

hub/test/src/@types/tape-promise/index.d.ts
```diff
@@ -0,0 +1,220 @@
+declare module 'tape-promise/tape' {
```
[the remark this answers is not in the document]
zone117x: Yeah, we actually just need to get this committed to https://github.com/DefinitelyTyped/DefinitelyTyped and eventually this will be a type package dependency. Its on my TODO list.

jcnelson:
Regarding test coverage, I see that we're only covering 37.5% of branches in `hub/server/revocations.ts`.  Should this be higher?

hub/src/server/authentication.ts
kantai: Yeah, this works the same as before (the auth unit tests would be failing if it changed).

kantai:
One question before merging to `develop` -- should we do a release of current `develop` branch before merging this? Current `develop` has revocation support and changes from the cloud integration testing PR in it.

zone117x:
@jude Good call. An important path in revocations wasn't being tested. Added integration tests for it and found a bug with revocations when using the `reader` sub-project's http server. Fixed in last commit.

kantai:
Cool, looks good to me, @zone117x -- to be clear, the previously untested path was fetching an uncached revocation number ?

zone117x:
@kantai 
> Cool, looks good to me, @zone117x -- to be clear, the previously untested path was fetching an uncached revocation number ?

That is correct. It is now being covered in integration tests against drivers. Only the gaia `reader` (used for the self-hosted DiskDriver integration test) ended up having a bug.

kantai:
Is this read to merge, @zone117x ? Since this fixes revocation paths currently in `develop`, we'll definitely want to merge this before releasing out of `develop`

zone117x:
Yeah good to go.

I pushed one last change the the nyc coverage config which was to enable code coverage reporting on all project source files. This drops total repo coverage to 77% due to source files inside the `reader` and `admin` projects that were previously untouched by tests and therefor not show up or count towards coverage. See full report at https://codecov.io/gh/blockstack/gaia/tree/88eaa8943a7a58cc92a0907d8e47a68ded9dda3e

This allows future PRs to add a bit of test coverage to a previously ignored file without causing a net decrease in repo coverage.

Issue #119, opened 2018-06-11
Gaia Hub should allow file deletes


kantai:
The gaia hub should allow file deletion. This will require an extension of the driver model, and a new endpoint (can use the `DELETE` HTTP method).

PR #214, opened 2019-03-13
Delete file feature

zone117x:
## Goal of PR

Implement delete file endpoint. 
https://github.com/blockstack/gaia/issues/119

## Implementation

Added support for deleting files using HTTP DELETE requests to the `/delete/${address}/${path}` endpoint. Authentication works the same way as on `/store`. Added new scopes `deleteFile` and `deleteFilePrefix` which work the same way as their `putFile` equivalents. 

On success, it returns a `202` status. If the path does not exist, is not a file, or is otherwise invalid, then a `403` status is returned. 

The `performDelete(...)` function is implemented in all drivers, with extensive tests for behavioral consistency. 

## Manual Testing

`npm test`

## Automated Testing

* Mock and integration tests implemented for all drivers. 
* http endpoint tests implemented.
* Server tests for auth / scoped auth / association token implemented. 

## Submission Checklist

      contains correct type annotations.


hub/src/server/drivers/AzDriver.ts
```diff
@@ -76,7 +76,9 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
       if (error.body && error.body.Code === 'ContainerAlreadyExists') {
         logger.info('Container initialized.')
       } else {
+        /* istanbul ignore next */
```
kantai: Why add these ignores?

README.md
```diff
@@ -326,6 +334,22 @@ document.
 
 ---
 
+```
+DELETE ${hubUrl}/delete/${address}/${path}
+```
+
+This performs a deletion of a file in the gaia hub at `${path}`. 
+
+On success, it returns a `202` status. Returns a `403` if the path does
```
jcnelson: Shouldn't this be a 404 if the path doesn't exist?  Also, 403 implies that the caller isn't authorized to do the action, not that the action is inapplicable to the object.  Should we instead be sending a 422 for the case where the object isn't a file, and a 400 for the case where the path itself is malformed?

hub/src/server/drivers/AzDriver.ts
```diff
@@ -180,6 +183,28 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
...
+      throw new BadPathError('Invalid Path')
+    }
+    const azBlob = `${args.storageTopLevel}/${args.path}`
+    const blobURL = azure.BlobURL.fromContainerURL(this.container, azBlob)
+    const blockBlobURL = azure.BlockBlobURL.fromBlobURL(blobURL)
+    try {
+      await blockBlobURL.delete(azure.Aborter.none)
+    } catch (error) {
+      if (error.statusCode === 404) {
+        throw new BadPathError('File does not exist')
```
jcnelson: This isn't an error due to the path being bad -- this is an error due to the file not existing.  Do we have a `DoesNotExist` exception somewhere that we can throw?

hub/src/server/drivers/GcDriver.ts
```diff
@@ -198,6 +199,30 @@ class GcDriver implements DriverModel, DriverModelTestMethods {
...
+    const filename = `${args.storageTopLevel}/${args.path}`
+    const bucketFile = this.storage
+      .bucket(this.bucket)
+      .file(filename)
+
+    try {
+      await bucketFile.delete()
+    } catch (error) {
+      if (error.code === 404) {
+        throw new BadPathError('File does not exist')
```
jcnelson: (same concern as in the Azure driver)

hub/src/server/drivers/S3Driver.ts
```diff
@@ -170,6 +175,35 @@ class S3Driver implements DriverModel, DriverModelTestMethods {
...
+
+    // S3 does not return an error if file does not exist, yet still writes a delete marker.
+    // So first check if the file exists using `headObject` which throws 404. 
+    // https://stackoverflow.com/a/53530749/794962
+    try {
+      await this.s3.headObject(s3params).promise()
+      await this.s3.deleteObject(s3params).promise()
+    } catch (error) {
+      if (error.statusCode === 404) {
+        throw new BadPathError('File does not exist')
```
jcnelson: (same concern as Azure driver)

hub/src/server/drivers/diskDriver.ts
```diff
@@ -184,26 +174,60 @@ class DiskDriver implements DriverModel {
...
   }
+
+  async performDelete(args: PerformDeleteArgs): Promise<void> {
+    const { absoluteFilePath, contentTypeFilePath } = this.getFullFilePathInfo(args)
+    let stat: fs.Stats
+    try {
+      stat = await fs.stat(absoluteFilePath)
+    } catch (error) {
+      if (error.code === 'ENOENT') {
+        throw new BadPathError('File does not exist')
```
jcnelson: (same concern as Azure driver)

hub/src/server/drivers/AzDriver.ts
zone117x: Because otherwise there is a net-decrease in coverage percentage and codecov (rightly) fails the build check. 
The alternative is to perform even deeper mocking into 3rd party libs to (very) artificially cause these edge-case error handlers to be reached. Not only is this level of mocking is an anti-pattern, but it will also discourage devs from writing defensive error handling code in general - which is already something we should be trying to move closer towards.

hub/src/server/http.ts
```diff
@@ -82,6 +82,37 @@ export function makeHttpServer(config: MakeHttpServerConfig & HubServerConfig &
...
+      filename = filename.substring(0, filename.length - 1)
+    }
+    const address = req.params[0]
+
+    server.handleDelete(address, filename, req.headers)
+      .then(() => {
+        res.writeHead(202)
+        res.end()
+      })
+      .catch((err: any) => {
```
jcnelson: Can we have distinct error paths for:

* the file not existing?
* the path being invalid?
* the path not referring to something that can be deleted?

README.md
zone117x: This is mirroring the putFile responses which is a 403 on BadPathError. The deleteFile method in drivers uses much of the same path validation. 
I agree we should change to those status codes. Should update putFile as well?

hub/src/server/drivers/AzDriver.ts
zone117x: I'll implement one

hub/src/server/http.ts
zone117x: Will do, I think this just requires creating a new error type and http response for file not existing.

zone117x: Fixed!

README.md
zone117x: This should now be fixed throughout the PR (all driver, http response handler, and docs). Can you confirm?

hub/src/server/drivers/AzDriver.ts
kantai: okay, sounds reasonable to me

zone117x:
Ping @kantai @jcnelson 
PR for blockstack.js putFile is in review https://github.com/blockstack/blockstack.js/pull/632
Would be nice to get this reviewed in time for next hub update.

hub/src/server/server.ts
```diff
@@ -73,6 +73,51 @@ export class HubServer {
...
+    const deletePaths = []
+    for (let i = 0; i < scopes.length; i++) {
+      if (scopes[i].scope == 'deleteFilePrefix') {
+        deletePrefixes.push(scopes[i].domain)
+      } else if (scopes[i].scope == 'deleteFile') {
+        deletePaths.push(scopes[i].domain)
+      }
+    }
+
+    if (deletePrefixes.length > 0 || deletePaths.length > 0) {
```
kantai: This is fine for now, but before we implement something else that's going to add new scopes (say "append" with appendPaths and appendPrefixes), we would probably want to factor this and the putFile scopes check so that we don't have a bunch of duplicated logic.

hub/src/server/http.ts
jcnelson: Looks good to me!  Thanks for addressing this :+1:

README.md
jcnelson: Yup, looks good!

jcnelson:
Looks good to me!

Issue #186, opened 2019-01-18
Docker build of /hub fails in `develop`


kantai:
Upgrading our blockstack.js dependency pulls in secp25k-tiny -- which depends on Python to `npm install` -- but our Dockerfile is an alpine nodejs image, so it doesn't have Python installed.

@wileyj thinks we should address this in all our (potentially affected) Dockerfiles by creating our own image which derives from alpine-node and adds Python. That way, we won't have to repeat these changes for other repos that depend on blockstack.js.

wileyj:
I'm handling this in our builds currently by sed'ing the base image in Dockerfile to `blockstack/node:latest`. 
Builds using this method are succeeding, but we should determine if this is the route we want to take. I'm not currently tagging these builds as `latest` or `master`. 

The `blockstack/node:latest` image is just the `node:8-alpine` base image with a few additional deps required for secp25k-tiny.

PR #215, opened 2019-03-22
Patch dockerfile for develop. Fix logging configuration.

kantai:
## Goal of PR

This patches the Dockerfile for `/hub` to build correctly with the latest changes in `develop` (see #186)

This also fixes some of the configuration behavior in `config.ts` with respect to logging in non-json mode (the default), and with timestamps.

## Implementation

This installs the requisite python packages for the npm install (required through a secp256k dependency).

## Manual Testing

You can run:

```
cd hub
docker build .
```

And it will run successfully.

## Automated Testing

This will build on our automated build systems.

### DIFF
--- a/hub/src/server/config.ts
+++ b/hub/src/server/config.ts
@@ -141,14 +141,15 @@ export function getConfig() {
 
   const formats = [
     config.argsTransport.colorize ? winston.format.colorize() : null,
-    config.argsTransport.json ? winston.format.json() : null,
-    config.argsTransport.timestamp ? winston.format.timestamp() : null
+    config.argsTransport.timestamp ?
+      winston.format.timestamp({ format: 'YYYY-MM-DD HH:mm:ss' }) : null,
+    config.argsTransport.json ? winston.format.json() : winston.format.simple()
   ].filter(f => f !== null)
   const format = formats.length ? winston.format.combine(...formats) : null
 
   logger.configure({
     format: format,
-    transports: [new winston.transports.Console(config.argsTransport)] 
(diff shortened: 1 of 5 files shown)


kantai:
You can test the logging differences:

On current develop:
```
$ npm run start
> gaia-hub@2.5.0 start /src/hub
> ts-node src/index.ts

undefined
```

With these changes:
```
$ npm run start
> gaia-hub@2.5.0 start /src/hub
> ts-node src/index.ts

warn: Listening on port 5000 in development mode ```

File: hub/src/server/config.ts [hunk 1]
```diff
@@ -141,14 +141,15 @@ export function getConfig() {
 
   const formats = [
     config.argsTransport.colorize ? winston.format.colorize() : null,
-    config.argsTransport.json ? winston.format.json() : null,
-    config.argsTransport.timestamp ? winston.format.timestamp() : null
+    config.argsTransport.timestamp ?
```
zone117x: Winston lib had a ton of changes, nice fix here. This is just getting the output to match the previous version output right? 
Also, I have this code block copy-pasted in all three sub-projects. Not happy with it, but didn't want to introduce cross-project dependencies in the PR (aside from the unit tests where it seems to work well). 
Should we update this in the other projects, or make a fourth sub-project called `common` that we can put this kind of stuff in, then cross reference? Typescript seems okay with path traversals outside of the `src` dir, so we wouldn't need to hacky package.json `"common-utils": "file:../common-utils"` NPM import.

kantai: I'll just update the other projects for now. We should probably pull things like this into a common sub-project at some point though.

PR #217, opened 2019-03-25
FileRename & FileStat operation primitives

zone117x:
## Goal of PR

The current collection proposal requires file rename and file stat capability. This is first pass implementation. 

These driver methods will likely only be used internally during the new collection putFile operation. 

## Implementation

S3 and Azure APIs do not support a rename/move operation. These are commonly requested features on both platforms. The recommended approach is to perform a copy then delete. Their CLI tools and other language libraries (like Java and C#) that implement anything like a `move` all do this copy & delete under the hood, on the client.

It appears a copy of a blob item within the same bucket is just as cheap as a proper rename (probably copy-on-write mechanisms being used). More research and citations needed. 

Examples:
* The Java S3 driver provides a `move` method which internally does a copy & delete: https://jets3t.s3.amazonaws.com/api/org/jets3t/service/StorageService.html#moveObject-java.lang.String-java.lang.String-java.lang.String-org.jets3t.service.model.StorageObject-boolean-

## Automated Testing

* Integration tests for all drivers for the file rename and file stat operations.
* Behavioral consistency tests for all cloud providers, including error conditions, for file rename and file stat. 

## Submission Checklist


markmhx:
I see that @yknl is currently set as a reviewer here even though it's WIP.

@zone117x do you want his eyes on the current code changes or should we remove him as a reviewer until it's no longer WIP?

Is this also still in progress or should we move it to the backlog until you have more time to resume work?

yknl:
This is in progress as part of the collections project, it will take some time to fully test and complete.

hub/src/server/authentication.ts
```diff
@@ -126,11 +128,16 @@ export class V1Authentication {
     // associationToken and verifies that it authorizes the "outer"
     // JWT's address (`bearerAddress`)
 
-    let associationToken: TokenType
-    try {
-      associationToken = decodeToken(token)
-    } catch (e) {
-      throw new ValidationError('Failed to decode association token in JWT')
+    const associationToken = (() => {
```
jcnelson: Should this snippit -- the `decodeToken(token)`, its try/catch block, and the subsequent `typeof` check -- be folded into its own method?  I see it's used multiple times.

hub/src/server/drivers/GcDriver.ts
```diff
@@ -222,7 +222,71 @@ class GcDriver implements DriverModel, DriverModelTestMethods {
...
+      }
+      /* istanbul ignore next */
+      logger.error(`failed to stat ${filename} in bucket ${this.bucket}`)
+      /* istanbul ignore next */
+      throw new Error('Google cloud storage failure: failed to stat ' +
+        ` ${filename} in bucket ${this.bucket}: ${error}`)
+    }
+  }
+
+  async performRename(args: PerformRenameArgs): Promise<void> {
```
jcnelson: On some of the `performRename` implementations, we follow up the rename with a `performStat()`-like operation on the new path to verify that the rename went through.  I'm guessing this is because rename isn't atomic (or even available as a first-class operation) on some storage providers?  If so, can we explicitly confirm whether or not each of these backends does renames atomically, and add a comment that points to the relevant documentation indicating as such?

hub/src/server/authentication.ts
zone117x: Fixed in https://github.com/blockstack/gaia/pull/245/commits/a792184381884306db0c47ee6b4c2b7d24856213
(This PR will have all package update related diffs removed once that PR is merged).

hub/src/server/drivers/GcDriver.ts
zone117x: S3 and Azure APIs do not support a rename/move operation. The recommended approach is to perform a copy then delete. Their CLI tools and other language libraries (like Java and C#) that implement anything like a `move` all do this `copy & delete` behavior under the hood, on the client. 

Integration tests for all providers validating `copy & delete` === `rename` for relevant drivers implemented in https://github.com/blockstack/gaia/pull/246/commits/f77e09db3ebfc8fb30f5d49c93a2c189813f3671

For concurrency issues -- a mutex around file ends points (`address/filename`) will be implemented around the http handler layer. Multi-hub load balancers will need to partition requests based on something like path prefix or hashing, rather than round-robin.

zone117x:
Updated the README driver model interface docs for the new operations.
Ping @jcnelson

hub/src/server/driverModel.ts
```diff
@@ -20,10 +20,31 @@ export interface PerformDeleteArgs {
...
+export interface StatResult {
+  exists: boolean;
+  lastModifiedDate?: number;
+  contentLength?: number;
+  contentType?: string;
+}
+
+export interface PerformRenameArgs {
+  path: string;
+  storageTopLevel: string;
```
jcnelson: Why does a rename allow for renaming files across different top-level paths?

hub/src/server/drivers/GcDriver.ts
jcnelson: Gotcha.  Thanks for addressing this!

hub/src/server/driverModel.ts
jcnelson: More specifically, does rename support the notion of moving a file between buckets or prefixes?  Asking because if so, even allowing this sounds very dangerous.

hub/src/server/drivers/AzDriver.ts
```diff
@@ -208,6 +208,75 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
...
+      }
+      /* istanbul ignore next */
+      logger.error(`failed to stat ${azBlob} in ${this.bucket}: ${error}`)
+      /* istanbul ignore next */
+      throw new Error('Azure storage failure: failed to stat' +
+        ` ${azBlob} in container ${this.bucket}: ${error}`)
+    }
+  }
+
+  async performRename(args: PerformRenameArgs): Promise<void> {
```
jcnelson: I'm not saying that we need to address this just yet, but in all drivers that don't support a direct `rename`, including this one, there's a slim chance that the Gaia hub could crash while it's in the middle of a rename.  This would leak space -- it would leave behind a copy of the data that it meant to delete.  What are the security and correctness implications of this, considering that rename is used in collections?

hub/src/server/driverModel.ts
zone117x: Using only one `storageTopLevel` arg makes sense to me. Will update that.

To give some context -- all the drivers just concat the these two strings together, and the relationship between end user buckets and this `storageTopLevel` is pretty implicit. The http handler and auth layers are currently responsible for enforcing anything like this, e.g. preventing cross-bucket file moves.

hub/src/server/drivers/AzDriver.ts
zone117x: This should just result in an additional "archival" file that is a duplicate of the canonical file. The client initiating the request (blockstack.js or the browser) would end up with an http error or timeout, and any proceeding collection-related operations *should* be aborted and the whole operation would need retried at some other point. 

I don't think this would cause any other issues, will be good to keep in mind during further collection implementation though.

PR #229, opened 2019-04-23
Schema validation warnings

zone117x:
## Goal of PR

From https://github.com/blockstack/gaia/pull/204#pullrequestreview-223293778

> While recently setting up a personal gaia hub I ran into a couple hiccups that required briefly going through the config related code. I think Hank just ran into similar situation. I think config schema checking that showed console warnings would be useful for this

## Implementation

Loosely based off original PR https://github.com/blockstack/gaia/pull/204

Uses the most popular json schema validation lib [`ajv`](https://www.npmjs.com/package/ajv). 

Outputs warnings for:
* Missing config params - only a couple are required by the schema right now. 
* Params that are of the wrong type or out of range. 
* (Most useful feature) "Extra" params found in the config, which are almost always typos or similarly accidental incorrect usage. 

### Config Warning Examples

```json
{
  "driver": "azure",
  "port": 12345,
  "whitelist": { "1Nw25PemCRv24UQAcZdaj4uD11nkTCWRTE": true }
}
```
\> `Config schema validation warning: config.whitelist should be array`

---

```json
{
  "driver": "microsoft",
  "port": 12345
}
```
\> `Config schema validation warning: config.driver should be equal to one of the allowed values`

---

```json
{
  "driver": "microsoft",
  "port": 3837659,
  "bucketTitle": "mybucket"
}
```
\> `Config schema validation warning: config['bucketTitle'] is an invalid additional property, config.driver should be equal to one of the allowed values, config.port should be <= 65535`

## Automated Testing

Full test coverage for schema validation warnings. 

## Submission Checklist

### DIFF
--- a/hub/config-schema.json
+++ b/hub/config-schema.json
@@ -157,15 +157,18 @@
         },
         "serverName": {
             "default": "gaia-0",
+            "description": "Domain name used for auth/signing challenges. \nIf `requireCorrectHubUrl` is true then this must match the hub url in an auth payload.",
             "type": "string"
         },
         "validHubUrls": {
+            "description": "If `requireCorrectHubUrl` is true then the hub specified in an auth payload can also be\ncontained within in array.",
             "items": {
                 "type": "string"
             },
             "type": "array"
         },
         "whitelist": {
+            "description": "List of ID addresses allowed to use this hub. Specifying this makes the hub private \nand only accessible to the specified addresses. Leaving this unspecified makes the hub \npublicly usable by any ID.",
             "items": {
                 "type": "string"
             },
--- a/hub/config-schema.json
+++ b/hub/config-schema.json
@@ -1,7 +1,9 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
+    "additionalProperties": false,
     "properties": {
         "argsTransport": {
+            "additionalProperties": false,
             "properties": {
                 "colorize": {
                     "default": true,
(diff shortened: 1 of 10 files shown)


markmhx:
@zone117x Is this still in progress or has worked paused? If the latter, we should move back to backlog or icebox based on whether it's still a short-term priority.

zone117x:
It's finished / ready for review

markmhx:
Great, who should review then? @kantai and @jcnelson?

hub/config-schema.json
```diff
@@ -1,7 +1,9 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
+    "additionalProperties": false,
```
jcnelson: Do we want to be this strict?  What's the downside of allowing anything?  I could see this becoming a source of compatibility problems down the road.

zone117x: None of this is strict because its console warnings. I've found this to actually be the most useful warning after setting up a config with something like `server_name` vs `serverName`.

zone117x: Note: I don't think we'd ever end up making this fully strict because some of the cloud credentials are very free-form

jcnelson: Ah, so to be clear -- if I add extra properties and use `Ajv` to validate the config against this schema, this will only result in a warning?  i.e. the program will otherwise continue to run?

zone117x: Yeah that is correct, only outputs to `console.warn`

PR #247, opened 2019-06-27
Implement mutex around file mutation http handlers

zone117x:
## Goal of PR

Implement mutex around file http handlers that perform file mutations: `POST /store/address/file` and `DELETE /delete/address/file`.

## Implementation

A mutex is unique to a `address/file` endpoint. When an operation is in progress for a given resource, subsequence requests to the modify the resource are immediately rejected with an HTTP 409 Conflict. 

## Manual Testing

Instantiating many concurrent POST or DELETE requests to the same endpoint, using something like curl or a load tester. 

## Automated Testing

Tests implemented for full coverage of the mutex implementation and http conflict error paths. 

## Submission Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,8 +248,13 @@ can be configured via the [config.json](#configuration-files).
 # Driver model
 
 Gaia hub drivers are fairly simple. The biggest requirement is the ability
-to fulfill the _write-to/read-from_ URL guarantee. As currently implemented
-a gaia hub driver must implement the following two functions:
+to fulfill the _write-to/read-from_ URL guarantee. 
+
+A driver can expect that each of its functions will be mutexed by a path. 
+As in, no function will be invoked concurrently with a given unique path. 
+
+As currently implemented
+a gaia hub driver must implement the following functions:
 
 ```javascript
 /**
--- a/README.md
+++ b/README.md
@@ -250,8 +250,8 @@ can be configured via the [config.json](#configuration-files).
 Gaia hub drivers are fairly simple. The biggest requirement is the ability
 to fulfill the _write-to/read-from_ URL guarantee. 
 
-A driver can expect that each of its functions will be mutexed by a path. 
-As in, no function will be invoked concurrently with a given unique path. 
+A driver can expect that two modification operations to the same path will be mutually exclusive. 
+No writes, renames, or deletes to the same path will be concurrent.
 
 As currently implemented
 a gaia hub driver must implement the following functions:
--- a/hub/src/server/http.ts
+++ b/hub/src/server/http.ts
@@ -80,13 +80,18 @@ export function makeHttpServer(config: HubConfigInterface): { app: express.Appli
       }
     }
 
-    if (!asyncMutex.tryAcquire(endpoint, handleRequest)) {
-      const errMsg = `Concurrent operation (store) attempted on ${endpoint}`
-      logger.error(errMsg)
-      writeResponse(res, { 
-        message: errMsg, 
-        error: errors.ConflictError.name 
-      }, 409)
+    try {
+      if (!asyncMutex.tryAcquire(endpoint, handleRequest)) {
+        const errMsg = `Concurrent operation (store) attempted on ${endpoint}`
+        logger.error(errMsg)
+        writeResponse(res, { 
+          message: errMsg, 
+          error: errors.ConflictError.name 
+        }, 409)
+      }
+    } catch (err) {
+      logger.error(err)
+      writeResponse(res, { message: 'Server Error' }, 500)
     }
 
   })
@@ -125,14 +130,20 @@ export function makeHttpServer(config: HubConfigInterface): { app: express.Appli
       }
     }
 
-    if (!asyncMutex.tryAcquire(endpoint, handleRequest)) {
-      const errMsg = `Concurrent operation (delete) attempted on ${endpoint}`
-      logger.error(errMsg)
-      writeResponse(res, { 
-        message: errMsg, 
-        error: errors.ConflictError.name 
-      }, 409)
+    try {
+      if (!asyncMutex.tryAcquire(endpoint, handleRequest)) {
+        const errMsg = `Concurrent operation (delete) attempted on ${endpoint}`
+        logger.error(errMsg)
+        writeResponse(res, { 
+          message: errMsg, 
+          error: errors.ConflictError.name 
+        }, 409)
+      }
+    } catch (err) {
+      logger.error(err)
+      writeResponse(res, { message: 'Server Error' }, 500)
     }
+
   })
 
   app.post(
--- a/hub/src/server/http.ts
+++ b/hub/src/server/http.ts
@@ -5,7 +5,7 @@ import cors from 'cors'
 import { ProofChecker } from './ProofChecker'
 import { getChallengeText, LATEST_AUTH_VERSION } from './authentication'
 import { HubServer } from './server'
-import { getDriverClass, logger } from './utils'
+import { getDriverClass, logger, AsyncMutexScope } from './utils'
 import { DriverModel } from './driverModel'
 import * as errors from './errors'
 import { HubConfigInterface } from './config'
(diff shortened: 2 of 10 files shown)


hub/src/server/utils.ts
```diff
@@ -63,3 +62,48 @@ export function timeout(milliseconds: number): Promise<void> {
...
+
+export interface AsyncMutex {
+  id: string;
+  owner: Promise<void>;
+  alive: boolean;
+}
+
+export class AsyncMutexScope {
+
+  private readonly _opened: Map<string, AsyncMutex> = new Map()
```
kantai: Why use AsyncMutex instead of using a `Set<string>` that only stores values if the mutex is currently held?

zone117x: `AsyncMutex` is just a lightweight placeholder object with possibly useful fields (currently not used since everything is handled in the `tryAcquire` method). Also thought `Map` was more performant than `Set<T>` but looks like that is not the case. Will simplify  👍

zone117x: Addressed in https://github.com/blockstack/gaia/pull/247/commits/7899979446a0818c925c3b75e94de710c1224990

hub/src/server/http.ts
```diff
@@ -74,7 +77,18 @@ export function makeHttpServer(config: HubConfigInterface): { app: express.Appli
...
+      }
+    }
+
+    if (!asyncMutex.tryAcquire(endpoint, handleRequest)) {
+      const errMsg = `Concurrent operation (store) attempted on ${endpoint}`
+      logger.error(errMsg)
+      writeResponse(res, { 
+        message: errMsg, 
+        error: errors.ConflictError.name 
+      }, 409)
```
jcnelson: Philosophical question -- should we be returning HTTP 409 or HTTP 429 here?  I remember saying 409 at one point since the client is (quite literally) making a conflicting request.  However, would it be better to advise the client on what to do here, instead of telling them what's wrong?  The advisable thing to do is to wait for a bit and try again, which would be indicated by HTTP 429.  Thoughts?

kantai:
I think we should update the README file describing the driver interface with information on the mutexes -- in particular that a driver can expect that each of its functions will be mutexed by path.

hub/src/server/http.ts
zone117x: I think HTTP 429 should be reserved for throttling or similar. The different error codes could inform the client on what retry/backoff/conflict-resolution strategy to attempt.

zone117x:
@kantai Readme for drivel model updated with all new interface functions and note about path mutex.
Ping @jcnelson for review.

README.md
```diff
@@ -248,8 +248,13 @@ can be configured via the [config.json](#configuration-files).
 # Driver model
 
 Gaia hub drivers are fairly simple. The biggest requirement is the ability
-to fulfill the _write-to/read-from_ URL guarantee. As currently implemented
-a gaia hub driver must implement the following two functions:
+to fulfill the _write-to/read-from_ URL guarantee. 
+
+A driver can expect that each of its functions will be mutexed by a path.
```
jcnelson: Grammar -- I don't think mutex is a verb?

"A driver can expect that two POSTs to the same path will be mutually exclusive.  No POSTs to the same path will be concurrent."

hub/src/server/http.ts
```diff
@@ -74,7 +77,18 @@ export function makeHttpServer(config: HubConfigInterface): { app: express.Appli
         } else {
           writeResponse(res, { message: 'Server Error' }, 500)
         }
-      })
+      }
+    }
+
+    if (!asyncMutex.tryAcquire(endpoint, handleRequest)) {
```
jcnelson: I notice that we _aren't_ guarding `list-files`, `revoke-all`, etc. with a per-path mutex.  Do we want to be doing this?

hub/src/server/utils.ts
```diff
@@ -63,3 +62,36 @@ export function timeout(milliseconds: number): Promise<void> {
...
+  /**
+   * If no mutex of the given `id` is already taken, then a mutex is created and the 
+   * given promise is invoked. The mutex is released once the promise resolves -- either by 
+   * success or error. 
+   * @param id A unique mutex name used in a Map.
+   * @param spawnOwner A function that creates a Promise if the mutex is acquired. 
+   * @returns `true` if the mutex was acquired, otherwise returns `false`
+   */
+  public tryAcquire(id: string, spawnOwner: () => Promise<void>): boolean {
+    if (this._opened.has(id)) {
```
jcnelson: We're _absolutely certain_ that this check will be atomic?

hub/src/server/utils.ts
```diff
@@ -63,3 +62,36 @@ export function timeout(milliseconds: number): Promise<void> {
...
+      return false
+    }
+
+    // Wrap in Promise.resolve to ensure potential synchronous errors are not throw within this function. 
+    const owner = Promise.resolve().then(() => spawnOwner())
+    this._opened.add(id)
```
jcnelson: This looks racey to me.  Is this execution schedule possible?  I.e. can the instantiation of `owner` in one coroutine cause it to yield to another coroutine trying to execute on the same path, such as below?  I suspect the answer will depend on implementation subtleties in node.js?

```
[Request 1]                    |[Request 2]
-------------------------------|-------------------------------------
tryAcquire("/foo")             |
if (this._opened.has(id)) {    |
   return false;               |
}                              |
                               |
const owner = Promise          |
  .resolve()                   |
  .then(() => spawnOwner())    |
                               |
<sleep>                        | <wakeup>
                               | tryAcquire("/foo")
                               | if (this._opened.has(id)) {
                               |   return false;
                               | }
                               |
                               | const owner = Promise
                               |  .resolve()
                               |  .then(() => spawnOwner()
<wakeup>                       | <sleep>
this._opened.add(id)           |
owner.finally(...              |
<sleep>                        | <wakeup>
                               | this._opened.add(id)  /* PROBLEM! */

File: README.md (hunk 3 above)
Comment (zone117x) [reply README.md]: Fixed

File: hub/src/server/http.ts (hunk 4 above)
Comment (zone117x) [reply hub/src/server/http.ts]: My feeling is no. `list-files` is readonly and should never be to the same path as a resource endpoint. Not sure about `revoke-all`, I don't think it would be used frequently enough to cause concurrency issues but perhaps possible. What do you think?

File: hub/src/server/utils.ts (hunk 6 above)
Comment (zone117x) [reply hub/src/server/utils.ts]: Good catch. The `.then` operation, even if synchronous, adds the inner invocation to the end of the event loop. Updated and double checked with some assertions:

```js
    let test = 1

    const runSpawn = () => {
      test++;
      return spawnOwner()
    }

    const owner = runSpawn()

    assert.equal(test, 2);
    this._opened.add(id)
```

The mutex acquirer was updated to handle potential synchronous exceptions that the `runSpawn()` promise may throw.

hub/src/server/utils.ts
zone117x: Addressed in below comment. I just went through the lock path again and everything is normal synchronous js, and nothing else resulting in event loop ordering issues.

hub/src/server/utils.ts
jcnelson: I don't think this fixes it?  Why is it still this:

```
const owner = spawnOwner()
this._opened.add(id)
```

Instead of this?

```
this._opened.add(id)
const owner = spawnOwner()
```

jcnelson: In the latter case, there's _no way_ the `spawnOwner()` call can trigger a coroutine yield.

kantai: I think that re-ordering could cause a dead lock if `spawnOwner` throws an exception.

jcnelson: So we'd want to wrap `spawnOwner()` in a try/catch/finally block where the `finally` branch de-registers `id`?

zone117x: The original race condition was because `spawnOwner()` was invoked within a `.then(...)` callback which caused it to execute _after_ the proceeding `this._opened.add(id)` statement due to event-loop behavior of promise chaining.

With the current code, the `spawnOwner()` statement is guaranteed to execute all of its _synchronous_ code before the `this._opened.add(id)` statement executes. 

However, there is a potential problem if callers attempt reentrant locking. That can be fixed by re-order the statements. The downside is having to re-throw the error within the catch statement after de-registering.

zone117x: https://github.com/blockstack/gaia/pull/247/commits/a7f04dad4d182310a927999b49f1997d5af1ab54 fixes the potential synchronous reentrancy issue if this class gets misused later. Uses a try/catch/re-throw to deal with synchronous errors.

hub/src/server/http.ts
jcnelson: I think it's fine for now, since the authenticator is the only client that will call this method.  We can fix it later if it proves to be a problem.

hub/src/server/utils.ts
jcnelson: LGTM

Issue #240, opened 2019-06-12
Collections: copy-write scope support


yknl:
In order to provide the copy/rename write operation used by collections, Gaia needs to support a new copy-write scope.

zone117x:
PRs merged into develop. Can we close or should we wait until deployed to master & prod?

yknl:
Let's close since it's been merged.

Issue #220, opened 2019-03-27
Collections: Support event-log write operation where existing file is renamed and history is kept


yknl:
To be completed

markmhx:
@zone117x will work with @yknl to update this issue with the latest thinking

markmhx:
@zone117x @yknl Is one of you actually working on this at the moment or should we move it to the backlog for now? It's been "in progress" for a month now.

yknl:
This is currently in progress

PR #248, opened 2019-06-28
putFileArchival

zone117x:
## Goal of PR

https://github.com/blockstack/gaia/issues/240, https://github.com/blockstack/gaia/issues/220

Implement the `putFileArchival` scope and historical file renaming scheme, enforced in putFile and deleteFile. 

## Implementation

When the scope is specified, file writes are not permitted to overwrite existing files. A file rename is performed using a historical naming scheme:
`{address}/foo/bar/photo.png` -> 
`{address}/foo/bar/.history.{timestamp}.{guid}.photo.png`

*Note: this is using the popular and fast [nanoid](https://zelark.github.io/nano-id-cc/) lib to generate a 10 char mixed-case-alphanumeric string rather than an actual uuid/guid.*

File deletes simply perform the same rename and nothing more. This results in the original file 404ing on request like a regular delete. 

## Discussion

This historical naming scheme differs from the current collections proposal in that it uses unix timestamps (milliseconds since epoch) rather than a contiguous incrementing version integer. 

This has the benefit of not depending on an index file and the additional read->write overhead, and does not require additional locks or expensive index file conflict checking & resolution. 

However, this does present the usual edge-cases with using timestamps:
* Severe server clock drift could result in incorrect file ordering. However, future file history restoration tooling could easily get around this by using the fileStat operation which returns the cloud driver provided `lastModifiedDate`. 
* Rapid repeated file writes to the same endpoint in conjunction with severe server clock drift could result in identical millisecond timestamps and result in historical file overwriting. Not sure if this is possible in practice because..
  1. Operations on an individual file are now serialized -- concurrent requests are rejected. So a second request would have to arrive exactly at the completion of the last request at the same moment that a clock glitch triggered a backward time drift. 
  2. A reachable Gaia hub necessitates an internet connection which means NTP should prevent severe clock drift?

If a timestamp cannot be used then I have a couple alternative proposals that also avoid bucket-wide index file race conditions and conflict resolution:

#### Per-file version tracking

Use a per-file historical version tracker rather than a bucket-wide index file. Something like:
```
myphoto1.jpg <----- Always the latest version
.history.status.myphoto1.txt <---- File containing the latest version integer
.history.1003.myphoto1.jpg <----- Previous version
.history.1002.myphoto1.jpg
.history.1001.myphoto1.jpg
```

The storage overhead of an additional small file seems equal to or better than the storage overhead of an index file containing a map of file paths and version integers. 

#### Rename with GUID

Use a randomly generated ID rather than an incrementing integer or timestamp. Future file history restoration tooling could use the cloud driver provided `lastModifiedDate` that is now available in the `fileStat` operation. This field has integration testing showing accuracy and behavioral consistency across all currently supported cloud storage providers. 

Thoughts?

## Manual Testing

Requires crafting a `putFileArchival` scoped token manually since it's not yet supported in the Browser or blockstack.js.

## Automated Testing

Full test coverage of new logic. 

## Submission Checklist

### DIFF
--- a/hub/src/server/authentication.ts
+++ b/hub/src/server/authentication.ts
@@ -56,9 +56,6 @@ export class AuthScopeValues {
 }
 
 export class AuthScopesTypes {
-  private constructor() {
-    // Disable class construction
-  }
   static readonly putFile = 'putFile'
   static readonly putFilePrefix = 'putFilePrefix'
   static readonly deleteFile = 'deleteFile'
--- a/hub/src/server/authentication.ts
+++ b/hub/src/server/authentication.ts
@@ -73,6 +73,21 @@ export function getTokenPayload(token: import('jsontokens/lib/decode').TokenInte
   return token.payload
 }
 
+export function decodeTokenForPayload(opts: { 
+  encodedToken: string; 
+  caller: string; 
+  validationErrorMsg: string; 
+  errorLogger: (msg: string) => void 
+}) {
+  try {
+    return getTokenPayload(decodeToken(opts.encodedToken))
+  } catch (e) {
+    opts.errorLogger(`${opts.caller}: ${opts.validationErrorMsg}, ${e}`)
+    opts.errorLogger(opts.encodedToken)
+    throw new ValidationError(opts.validationErrorMsg)
+  }
+}
+
 export interface AuthenticationInterface {
   checkAssociationToken(token: string, bearerAddress: string): void
   getAuthenticationScopes(): AuthScopeEntry[]
@@ -100,15 +115,12 @@ export class V1Authentication implements AuthenticationInterface {
       throw new ValidationError('Authorization header should start with v1:')
     }
     const token = authPart.slice('v1:'.length)
-    const payload = (() => {
-      try {
-        return getTokenPayload(decodeToken(token))
-      } catch (e) {
-        logger.error(e)
-        logger.error('fromAuthPart')
-        throw new ValidationError('Failed to decode authentication JWT')
-      }
-    })()
+    const payload = decodeTokenForPayload({
+      encodedToken: token, 
+      caller: 'fromAuthPart', 
+      validationErrorMsg: 'Failed to decode authentication JWT',
+      errorLogger: logger.error
+    })
 
     const publicKey = payload.iss
     if (!publicKey) {
@@ -173,13 +185,12 @@ export class V1Authentication implements AuthenticationInterface {
     // associationToken and verifies that it authorizes the "outer"
     // JWT's address (`bearerAddress`)
 
-    const payload = (() => {
-      try {
-        return getTokenPayload(decodeToken(token))
-      } catch (e) {
-        throw new ValidationError('Failed to decode association token in JWT')
-      }
-    })()
+    const payload = decodeTokenForPayload({
+      encodedToken: token, 
+      caller: 'checkAssociationToken', 
+      validationErrorMsg: 'Failed to decode association token in JWT',
+      errorLogger: logger.error
+    })
 
     // publicKey (the issuer of the association token)
     // will be the whitelisted address (i.e. the identity address)
@@ -235,15 +246,13 @@ export class V1Authentication implements AuthenticationInterface {
    * Returns [] if there is no association token, or if the association token has no scopes
    */
   getAuthenticationScopes() {
-    const payload = (() => {
-      try {
-        return getTokenPayload(decodeToken(this.token))
-      } catch (e) {
-        logger.error(this.token)
-        logger.error('getAuthenticationScopes')
-        throw new ValidationError('Failed to decode authentication JWT')
-      }
-    })()
+
+    const payload = decodeTokenForPayload({
+      encodedToken: this.token, 
+      caller: 'getAuthenticationScopes', 
+      validationErrorMsg: 'Failed to decode authentication JWT',
+      errorLogger: logger.error
+    })
 
     if (!payload.hasOwnProperty('scopes')) {
       // not given
@@ -281,15 +290,12 @@ export class V1Authentication implements AuthenticationInterface {
                         options?: { requireCorrectHubUrl?: boolean,
                                     validHubUrls?: Array<string>,
                                     oldestValidTokenTimestamp?: number }): string {
-    const payload = (() => {
-      try {
-        return getTokenPayload(decodeToken(this.token))
-      } catch (e) {
-        logger.error(this.token)
-        logger.error('isAuthenticationValid')
-        throw new ValidationError('Failed to decode authentication JWT')
-      }
-    })()
+    const payload = decodeTokenForPayload({
+      encodedToken: this.token, 
+      caller: 'isAuthenticationValid', 
+      validationErrorMsg: 'Failed to decode authentication JWT',
+      errorLogger: logger.error
+    })
 
     const publicKey = payload.iss
     const gaiaChallenge = payload.gaiaChallenge
--- a/hub/src/server/authentication.ts
+++ b/hub/src/server/authentication.ts
@@ -13,29 +13,61 @@ function pubkeyHexToECPair (pubkeyHex: string) {
   return bitcoinjs.ECPair.fromPublicKey(pkBuff)
 }
 
-export type AuthScopeType = {
-  scope: string,
+export interface AuthScopeEntry {
+  scope: string
   domain: string
 }
 
-export type TokenPayloadType = {
-  gaiaChallenge: string,
-  iss: string,
-  exp: number,
-  iat?: number,
-  salt: string,
-  hubUrl?: string,
-  associationToken?: string,
-  scopes?: AuthScopeType[],
+export interface TokenPayloadType {
+  gaiaChallenge: string
+  iss: string
+  exp: number
+  iat?: number
+  salt: string
+  hubUrl?: string
+  associationToken?: string
+  scopes?: AuthScopeEntry[]
   childToAssociate?: string
 }
 
-export const AuthScopes = [
-  'putFile',
-  'putFilePrefix',
-  'deleteFile',
-  'deleteFilePrefix'
-]
+export class AuthScopeValues {
+
+  writePrefixes: string[] = []
+  writePaths: string[] = []
+  deletePrefixes: string[] = []
+  deletePaths: string[] = []
+  writeArchivalPrefixes: string[] = []
+  writeArchivalPaths: string[] = []
+
+  static parseEntries(scopes: AuthScopeEntry[]) {
+    const scopeTypes = new AuthScopeValues()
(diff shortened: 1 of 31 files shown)


hub/src/server/authentication.ts
```diff
@@ -330,7 +379,17 @@ export class V1Authentication {
...
 
-export class LegacyAuthentication {
+export class LegacyAuthentication implements AuthenticationInterface {
+
+  checkAssociationToken(_token: string, _bearerAddress: string): void {
+    throw new Error('Method not implemented.')
+  }
+
+  parseAuthScopes(): AuthScopeValues {
+    // TODO: can legacy auth tokens be maliciously created to get around restrictive auth scopes?
```
zone117x: @kantai thoughts on this?

hstove:
I'm in favor of anything removing the read/write overhead. That includes using timestamps and no bucket-wide index file.

hub/src/server/authentication.ts
kantai: No -- I don't think they'd be able to get around auth scopes any more than someone that possesses the private key already could.

However, we should look to deprecate support for these auth tokens -- blockstack.js hasn't used them for about a year now.

jcnelson:
> This historical naming scheme differs from the current collections proposal in that it uses unix timestamps (milliseconds since epoch) rather than a contiguous incrementing version integer.

Speaking to this topic, you can avoid the need for a dedicated index file while simultaneously avoiding any clock-drift patterns by using both a timestamp and a GUID in the `.history` file name (i.e. your file name pattern would be `.history.{timestam}.{GUID}.{name}`).  Then, you would treat any writes that occur within the same millisecond as "simultaneous", and simply use their GUIDs'  lexical ordering to break ties.

kantai:
I'm in favor of this approach paired with a nonce (or a random integer):

 `.history.{timestamp}.{nonce}.{name}`

As unlikely as an overlap would be with a random integer, a local incrementing nonce would eliminate the possibility of overwrite entirely (and could be transient -- it doesn't need to live past a ms). I'm not sure how much that matters though -- if we used a random integer, the possibility of overwrite is very low.

zone117x:
Updated to use a `.history.{timestamp}.{GUID}.{name}` naming scheme. Moved PR out of draft and ready for review. 

Will update PR with more documentation on the changes if/when the implementation is reviewed and accepted.

hub/src/server/server.ts
```diff
@@ -74,6 +75,23 @@ export class HubServer {
     }
   }
 
+  getHistoricalFileName(address: string, filePath: string) {
+    const pathParts = filePath.split('/')
+    const fileName = pathParts[pathParts.length - 1]
+    const filePathPrefix = filePath.slice(0, filePath.length - fileName.length)
+  
+    // reject writes to historical files
+    if (fileName.startsWith('.history.')) {
```
jcnelson: I'm not sure why this check is here.  The file's name should be able to be anything, since we're going to prefix it with `.history.${Date.now()}.${generateUniqueID()}.`, right?

hub/src/server/server.ts
```diff
@@ -83,24 +101,31 @@ export class HubServer {
...
     // can the caller delete? if so, in what paths?
     const scopes = getAuthenticationScopes(requestHeaders.authorization)
-    const deletePrefixes = []
-    const deletePaths = []
-    for (let i = 0; i < scopes.length; i++) {
-      if (scopes[i].scope == 'deleteFilePrefix') {
-        deletePrefixes.push(scopes[i].domain)
-      } else if (scopes[i].scope == 'deleteFile') {
-        deletePaths.push(scopes[i].domain)
+    const isArchivalRestricted = scopes.writeArchivalPaths.length > 0 || scopes.writeArchivalPrefixes.length > 0
```
jcnelson: This code is duplicated below.  I think there could be a method that determines (1) whether or not we're authorized to write to this path and (2) if so, whether or not we're archival-restricted.

hub/test/src/testServer.ts
```diff
@@ -472,6 +474,136 @@ export function testServer() {
...
+
+      const getDataStream = () => {
+        const s = new PassThrough()
+        s.end('hello world')
+        return s
+      }
+
+      // test .history. file write rejection
+      try {
+        await server.handleRequest(testAddrs[0], 'baz/.history.not_ok.txt', {
```
jcnelson: I think this should be allowed -- we're going to prepend `.history.` ... to the file name anyway.

hub/src/server/server.ts
zone117x: There was a concern at one point to prevent restricted apps from deleting or zeroing out historical files, but that should not be possible with current implementation. Attempting to do so would just create a historical file for a historical file.

So, this check just prevents that historical-file-inception. I think this is would be more an accidental thing an app dev might do. For example listFiles, iterate, write to a historical file, say `.history.1564526417230.nTXJqe7gTc.myphoto.png`, and trigger a historical rename of the historical file: `history.1564526772589.C9H8GzbAlA..history.1564526417230.nTXJqe7gTc.myphoto.png`

But doesn't seem like that big of a deal. Should we allow it?

zone117x: This also brings up the question -- should the existence of a `putFileArchival` scope also restrict `list-files`? For example, not returning entries for any of the historical files. 

One oddity would be that a user's bucket may contain >PAGE_LIMIT historical files, and end up with lots of `list-file` responses that contain a pagination marker but no file entries.

jcnelson: > So, this check just prevents that historical-file-inception.

I don't think this is a concern.  A buggy app is a buggy app.

> This also brings up the question -- should the existence of a putFileArchival scope also restrict list-files? For example, not returning entries for any of the historical files.

At the very least, I think `list-files` should hide the timestamp unless it had `putFileArchival` permission, since this would allow an app to view the user's activity.  But I take your point about there being an inexplicably large number of empty pages.  Maybe `list-files` should censor the name and timestamp of history files so they all say `<hidden>` and have a timestamp of `0`?

hub/src/server/server.ts
zone117x: DRY'd this up

hub/src/server/server.ts
zone117x: @yknl Any thoughts on this?

kantai: I think on these two questions,

1) We should allow historical naming inception. It's not a safety concern, so why add this caveat to the data model (i.e., you can name your file anything but something starting with `.history.`).
2) Historical files _should_ appear in `list-files` for the time being. `list-files` isn't really usable on the typical read path for applications (because it doesn't really work for multi-player reads), but it's primary use is for supporting (a) backups, and, in the case of archival puts, (b) rollbacks. So, if an application wanted to actually find an archived version of a file, it could do so using `list-files`. If the files were hidden, that wouldn't be possible.

zone117x: The `.history.` name prefix restriction has been removed.

hub/test/src/testServer.ts
zone117x: Done

hub/src/server/server.ts
yknl: My understanding is that list-file with an archival scoped auth token should not return historical files. When the file manager in the browser is implemented, it can use a full scope auth token to view historical files.

zone117x: In the collections meeting we decided to not include historical files in list-files with an archival scope -- PR has been updated.

zone117x: The blockstack.js `listFiles` loop will break with empty pages due to historical file filtering https://github.com/blockstack/blockstack.js/blob/76a1fb23ffc7db6ebb371698057da8baa48e8434/src/storage/index.ts#L696

Not sure how to fix this cleanly without also updating blockstack.js. Gaia could detect this situation and return a single 'null' item in the entries array which would cause the listFile loop to continue, but probably break apps that are not expecting null entries. Thoughts?

kantai: I continue to think that the historical files should show up in list-files. What's the downside to this? Do applications use list-files for any functionality?

jcnelson: I think it's mainly a privacy concern -- an application can see the historic states of a collection file.

zone117x: @kantai Yeah some privacy concerns.

Yes apps do use listFiles:
https://github.com/search?utf8=%E2%9C%93&q=%22blockstack%22+%22listfiles%22&type=Code&ref=advsearch&l=&l=

Apps or blockstack.js would need to implement parsing & handling of the historical file naming scheme. This will probably need to be done at least in the Browser at some point anyway for the restoration tool, but that can be done later.

hub/src/server/server.ts
```diff
@@ -59,23 +61,47 @@ export class HubServer {
...
-      page: page
-    }
-    if (stat) {
-      const result = await this.driver.listFilesStat(listFilesArgs)
-      if (isArchivalRestricted) {
-        result.entries = result.entries.filter(entry => !this.isHistoricalFile(entry.name))
+    let attemptCount = 0
+    const maxFetchPageAttempts = this.config.fetchPageAttempts || 100
+    
+    const fetchPage = async (pageArg: string): Promise<ListFilesResult | ListFilesStatResult> => {
```
jcnelson: I disagree with this approach.  I think that `list-files` should just censor sensitive data, but it shouldn't make the Gaia hub do a bunch of background work like this.  This is just asking for a DDoS.  Instead, it should return a sentinel file name that indicates that the true name is obfuscated or will not be returned.

If the purpose of this code is to make it so blockstack.js's `listFiles()` doesn't return such files to the application, then it should be _blockstack.js_ that re-requests data.

kantai: Yep -- we want to make sure that we're doing something like a constant number of lookups for any given request, otherwise it's a denial of service vector.

zone117x: I don't really disagree, but clients will end up requesting the hub to do this same thing. I'm okay with blockstack.js being updated to handle the situation. 
@yknl I can do the blockstack.js PR for this -- does that sound good to you?

yknl: Yea, I think the logical solution here is to have blockstack.js handle empty pages instead of processing it on the hub. 

Just to re-iterate why we shouldn't return `.history` files to blockstack.js `listFile`:
1. It makes `listFiles` on collections almost useless due to the number of irrelevant historical files it returns. 
2. There's a privacy concern with exposing the historical files.

zone117x: PR updated to return a single `null` entry when this situation occurs. blockstack.js needs updated to A) filter these null results from the app and, B) continue to request additional pages.

zone117x: Corresponding blockstack.js PR: https://github.com/blockstack/blockstack.js/pull/685

PR #249, opened 2019-07-01
Option to return fileStat info during listFiles

zone117x:
## Goal of PR

Related to https://github.com/blockstack/gaia/issues/220, https://github.com/blockstack/gaia/issues/240

Provide ability to query file metadata (primarily `lastModifiedDate`) with the `listFiles` endpoint. 

This may be needed by a future Collections historical file restoration tool -- pending discussion on historical naming scheme in https://github.com/blockstack/gaia/pull/248:

> If a [gaiahub server] timestamp cannot be used [in the historical file name] then I have a couple alternative proposals that avoid bucket-wide index file race conditions and conflict resolution ... Use a random ID in this historical file name. Future file history restoration tooling could use the cloud driver provided `lastModifiedDate` that is now available in the fileStat operation. This field has integration testing showing accuracy and behavioral consistency across all currently supported cloud storage providers.

This feature does *not* cause additional request overhead with the current cloud providers -- their `list file` APIs already return this metadata alongside the filenames. The only additional overhead should be the additional data returned to clients that request it. 

## Implementation

Extends the existing `POST ${hubUrl}/list-files/${address}` endpoint. 
If the POST body contains a `stat: true` field then the returned JSON includes the fileStat metadata:

```js
{
  entries: [
    { name: string, lastModifiedDate: number, contentLength: number },
    { name: string, lastModifiedDate: number, contentLength: number },
    ...
  ]
  page: string // optional pagination marker
}
```

Without `stat: true` the previous behavior is preserved -- the returned JSON format remains:
```js
{
  entries: string[], // file name strings
  page: string // optional pagination marker
}
```

## Testing

Integration tests for all supported cloud providers -- all showing behavioral consistency. 

## Submission Checklist


jcnelson:
I'm in favor of this regardless of the eventual collections history naming scheme.

larrysalibra:
Agree with @jcnelson

Issue #253, opened 2019-08-01
Production ready node.js HTTP server


zone117x:
Meta ticket for the tasks required to remove an extra web server dependency (like nginx) as a requirement for running a gaia hub:

* [x] `/hub_info` should advertise the hub's maximum file size  https://github.com/blockstack/gaia/issues/161
* [x] Check content-size on gaia hub writes (in js http handler) https://github.com/blockstack/gaia/issues/187
* [x] Support SSL config in node.js https://github.com/blockstack/gaia/issues/252

jeffdomke:
@muneebm @radicleart @radiclesociety you mentioned Gaia setup perviously, just tagging you here so you can keep up to date. This has be prioritized.

zone117x:
This has been implemented and merged in dev branch since PR https://github.com/blockstack/gaia/pull/270 in September. 

There is a minor CORs bug, related to this feature, when running the write-endpoint node.js server without a reverse HTTP proxy, which is fixed in the soon to be merged PR https://github.com/blockstack/gaia/pull/276 

These lines, specifically: 
https://github.com/blockstack/gaia/pull/276/commits/395078027290dd01076d99365284c27cfcbfbfce#diff-e41f0e5dc3ecba29e76d994f3fcafb16R46-R60

Issue #187, opened 2019-01-18
Edit gaia hub drivers to check content-size on gaia hub writes 


GInxh:
chunks the upload progress for size monitoring.

markmhx:
@GInxh Does this issue pertain to the containers epic?

GInxh:
This effects production and develop branch, and also what I am working on for containerizing the develop branch.

GInxh:
@markmhx not apart of the epic since if effects master and develop branch and not just the http feature/docker-compose branch

markmhx:
@GInxh Is this issue strictly necessary to fulfill that epic even if it affects those other branches? 

If not, then I agree let's remove from the epic. But if so, I think it'd be good to include it.

kantai:
We've generally recommended enforcing something like this via a reverse proxy (e.g.) nginx. The advantage of doing that is that it ensures we don't create an extra in memory copy of the request body (which can be quite easy to do accidently).

If we wanted to add this functionality to the hub itself, we'd probably not want to do this via the drivers, because this functionality would apply to all of the drivers, and doesn't really depend on any driver-specific issues. We could add it to the handler in `hub/src/server/http.js` or `server.js`. The tricky bit is ensuring that you don't ever buffer the request's body. There's an express middleware class that could handle this pretty easily -- `body-parser`, but it buffers the request body. What we'd want is something that simply places a limit on the amount of data that can read out of a given stream.

zone117x:
This article has some good info on avoiding those types of problems https://nodejs.org/en/docs/guides/backpressuring-in-streams/#lifecycle-of-pipe
Looks like staying within the `pipe` API is key. The entire body can avoid being copied, but I think size enforcement requires monitoring chunks and those probably are copied/GC'd.

zone117x:
After some more experience with our driver implementations and the Nodejs stream/pipe API, this is really easy to implement and get right. It doesn't need implemented per driver, just one spot, and probably just using the [stream.PassThrough](https://nodejs.org/api/stream.html#stream_class_stream_passthrough) to count chunk bytes and `.destroy()` the stream when limit is exceeded.

timstackblock:
Tested as part of https://github.com/blockstack/gaia/pull/256

ghost:
@markmhendrickson 

I will be deleting my github and per github policy all code & comments in repositories I do not own will remain (production will not be impacted) and replaced with github general ghostuser account https://github.com/ghost

[...]

Issue #161, opened 2018-12-12
GET /hub_info should advertise the hub's maximum file size (if set)


jcnelson:
If the Gaia hub enforces a maximum file size, then it should advertise the size via `GET /hub_info` in a standard way.  This could be set in the config file (or an envar).

zone117x:
Closed by https://github.com/blockstack/gaia/pull/256

PR #256, opened 2019-08-06
Implement `maxUploadFileSize`

zone117x:
Progress towards [Production ready node.js HTTP server ](https://github.com/blockstack/gaia/issues/253)

Completes issues:
* https://github.com/blockstack/gaia/issues/161 GET /hub_info should advertise the hub's maximum file size 
* https://github.com/blockstack/gaia/issues/187 Edit gaia hub drivers to check content-size on gaia hub writes 

When a content-length header is provided (almost always the case) then it is checked against the hub's max configured size. If the size is too large, the http request is immediately rejected with a specific error for the client (blockstack.js) to be able to handle appropriately. 

Additionally, while the a request payload is being streamed, the size is monitored and if the content-length is exceeded, the stream is immediately destroyed. 

If a content-length header is not provided, then the max hub size limit is used instead. This is to allow for clients that do not necessarily know the content-length in advance. The `fetch` w3c spec specifies that a ReadableStream (which has an unknown length) should be allowed as the POST body payload. This is a feature we may want to use in the future for streaming encryption and network request. 

## Manual Testing

TODO: Memory load testing during large file writes to ensure there are no performance regressions. 

## Automated Testing

Near full test coverage. SuperAgent is a PITA and difficult to spoof malicious client behavior. Some error paths are not covered in the http code reporting layer, but covered in the error handling layer right beneath it. 

## Submission Checklist


hub/src/server/server.ts
```diff
@@ -240,7 +256,45 @@ export class HubServer {
...
+    const monitorPipeline = pipelineAsync(stream, monitorStream)
+
+    // Create a PassThrough stream to give to driver for uploading to storage backend. 
+    const uploadStream = new PassThrough()
+    const uploadPipeline = pipelineAsync(stream, uploadStream)
+
+    // Now that the PassThrough streams and pipes are setup, the data event listener
+    // for the size monitor can be setup without either stream listeners missing any chunks. 
+    let monitoredContentSize = 0
+    monitorStream.on('data', (chunk: Buffer) => {
```
zone117x: Reviewer note:
This creates two reader streams for a file upload. This has the potential, depending on node/V8 optimizations, to briefly duplicate network chunk streams in memory. But in general, this is the recommended approach to properly handling backpressure and monitoring stream progress when there are multiple consumers.

hub/src/server/server.ts
```diff
@@ -240,7 +256,29 @@ export class HubServer {
       }
     }
 
-    const readURL = await this.driver.performWrite(writeCommand)
+    // Create a PassThrough stream to monitor streaming chunk sizes. 
+    const { monitoredStream, pipelinePromise } = monitorStreamProgress(stream, totalBytes => {
+      if (totalBytes > this.maxFileUploadSizeBytes) {
```
kantai: This should check against the Content-Length header -- if the uploader is streaming more than the claimed Content-Length, we could close and error.

zone117x: Good catch, fixed.

jeffdomke:
Sounds like next steps are @jcnelson finishes review then @timstackblock will help on the memory  load testing.

hub/src/server/server.ts
```diff
@@ -208,14 +213,25 @@ export class HubServer {
...
-      path, stream, contentType,
-      contentLength: parseInt(requestHeaders['content-length'] as string)
+    await this.proofChecker.checkProofs(address, path, this.getReadURLPrefix())
+
+    const contentLengthHeader = requestHeaders['content-length'] as string
+    const contentLengthBytes = parseInt(contentLengthHeader)
+    const isLengthFinite = Number.isFinite(contentLengthBytes)
+
+    if (!isLengthFinite) {
+      const errMsg = `A valid 'Content-Length' header must be passed. Received header "${contentLengthHeader}"`
```
jcnelson: I think we should mention somewhere in the README that a `Content-Length` header is necessary -- there's no support for chunked encoding.

hub/src/server/utils.ts
```diff
@@ -43,6 +43,14 @@ export function getDriverClass(driver: DriverName): DriverConstructor & DriverSt
   }
 }
 
+export function megabytesToBytes(megabytes: number) {
+  return megabytes * 1024 * 1024
+}
+
+export function bytesToMegabytes(bytes: number, decimals = 2) {
+  return Number.parseFloat((bytes / 1024 / 1024).toFixed(decimals))
```
jcnelson: Not a bug, but I'd prefer `bytes / (1024 * 1024)` -- I think `/` is always left-binding, but this would play it safe.

timstackblock:
Confirmed the hub and reader is successfully running. Overall, the PR respects upper file limits but doesn’t respect lower limits. 

	•	3 sample files are needed: 
             file1 is less than 100KB, 
             file2 is greater than 100KB but less than upperlimit (20MB) default and 
             file3 is greater than upperlimit.

File1 upload (Successful):
	•	File upload was successful despite being less than the lower limit of 100KB.
 
File2 Upload(Successful):
	•	File upload was successful given it was within the range of acceptable file size limit.
 
File3 Upload(Unsuccessful):
	•	File size exceeds upper size limit.
	•	Response: "message":"Max file upload size is 20 megabytes. Rejected Content-Length of 109.4165 megabytes

hub/src/server/server.ts
zone117x: Per my comment https://github.com/blockstack/gaia/pull/256#issuecomment-525489836
I removed the requirement for providing a content-length header. After extensive testing, this is a frustrating header to provide given CORs whitelist contraints. Additionally, it excludes future uploading features that do not necessarily know the content length in advance.

hub/src/server/utils.ts
zone117x: Fixed

zone117x:
I removed the requirement for providing a content-length header. After extensive testing, this is a frustrating header to provide given CORs whitelist contraints. Additionally, it excludes future uploading features that do not necessarily know the content length in advance. 

If the header is provided by a well-behaving and non-streaming hub and client, then it results in nice immediate errors immediately provided to the client. If the header is not provided -- as is currently the case with the current PBC hub -- then the error is provided to the client after the hub-configured limit has been exceeded. 

I have pending future PRs that increase the probability of Gaia hub servers and clients reporting the content-length values. However, the possibility of the header not being provided is a situation in which I think we should reasonably expect and allow for. 

@jcnelson @kantai can you re-review? Thanks.

PR #257, opened 2019-08-13
Add "deploy on kubernetes" button example

pastudan:
## Goal of PR

Show example of KubeSail template button


moxiegirl:
@pastudan Thank you for your submission. Instead of adding this link to our repo README, we added this to the list of community resources in our documentation. That will give it wider exposure. You can find the list here:

https://docs.blockstack.org/develop/communityresources

PR #263, opened 2019-08-27
Merge develop into master: collections dependency

zone117x:
## Goal of PR

Merge develop into master, in preparation for updated PBC hub deployment. This primary contains changes related to collections. 

Copy & paste of the changelog:

### Added
- Implemented the `putFileArchival` restrictive auth scope which causes any 
  file modifications such as `POST /store/...` or `DELETE /delete/...` to 
  "backup" the original file by using a historical naming scheme. For example, 
  a file write to `{address}/foo/bar/photo.png` will cause the original
  file, if it exists, to be renamed to 
  `{address}/foo/bar/.history.{timestamp}.{guid}.photo.png`. 
- The `/list-files/${address}` endpoint now returns file metadata 
  (last modified date, content length) if the `POST` body contains 
  a `stat: true` option. 
- Implemented `readFile` and `fileStat` methods on all drivers, however, 
  these are not yet in use or publicly exposed via any endpoints. 
### Changed
- Concurrent requests to modify a resource using the `/store/${address}/...`
  or `/delete/${address}/...` endpoints are now always rejected with a 
  `HTTP 409 Conflict` error. Previously, this was undefined behavior
  that the back-end storage drivers dealt with in different ways. 
- The `Access-Control-Max-Age` header for preflight CORs OPTION responses
  set to 24 hours.


hub/src/server/authentication.ts
```diff
@@ -126,18 +185,18 @@ export class V1Authentication {
...
 
-    let associationToken: TokenType
-    try {
-      associationToken = decodeToken(token)
-    } catch (e) {
-      throw new ValidationError('Failed to decode association token in JWT')
-    }
+    const payload = decodeTokenForPayload({
+      encodedToken: token, 
+      caller: 'checkAssociationToken',
```
jcnelson: This structure seems odd to me.  Why can't this code simply throw an error here, instead of having to pass some error context into another function?

hub/src/server/authentication.ts
```diff
@@ -308,44 +369,54 @@ export class V1Authentication {
...
-export class LegacyAuthentication {
-  publickey: bitcoin.ECPair
+export class LegacyAuthentication implements AuthenticationInterface {
+
+  checkAssociationToken(_token: string, _bearerAddress: string): void {
+    throw new Error('Method not implemented.')
+  }
+
+  parseAuthScopes(): AuthScopeValues {
+    // TODO: can legacy auth tokens be maliciously created to get around restrictive auth scopes?
```
jcnelson: We need to know the answer to this before we merge!

jcnelson: The answer needs to be a definitive "no" -- legacy auth tokens must be assigned a restricted set of scopes that the caller would have expected to receive (e.g. the ability to store data, but not delete or list data or add to collections).

jcnelson: Sounds good.  We should remove this comment then.

jcnelson:
Agree with @kantai -- the README needs to be updated to speak to the constraints imposed on replicated hubs.

hub/src/server/authentication.ts
zone117x: Removed comment

zone117x:
Added a note to the README about deployments with replicated nodes.

jcnelson:
LGTM :+1:

PR #270, opened 2019-09-30
Https support

zone117x:
## Goal of PR

Implement support for the [Node.js https server](https://nodejs.org/docs/latest-v10.x/api/https.html) in order to simplify hub deployments. This makes it easier to orchestrate hub deployments for various compute providers and DNS/DDNS providers. 

## Implementation

Two options for enabling https are provided. See the updated `hub/config-schema.json` file diff for all supported config options. These are also all configurable via env vars. 

### Using TLS cert files

Supports both cert files in both the `PFX / PKCS12` format and the `PEM key & cert chain` format. 

Example of the config values for typical `pfx` cert file usage:
```json
{
  "enableHttps": "cert_files",
  "tlsCertConfig": {
    "pfxFile": "~/.config/ssl/cert.pfx"
  }
}
```

Example of the config values for the typical `PEM` files usage: 
```json
{
  "enableHttps": "cert_files",
  "tlsCertConfig": {
    "keyFile": "~/.config/ssl/key.pem",
    "certFile": "~/.config/ssl/cert.pem"
  }
}
```

### Using ACMEv2 client

This uses the [`greenlock-express`](https://www.npmjs.com/package/greenlock-express) middleware to provide support for Lets Encrypt v2 (i.e. [`ACME draft-12`](https://tools.ietf.org/html/draft-ietf-acme-acme-12)). 

The default ACME standard [challenge type `http-01`](https://letsencrypt.org/docs/challenge-types/#http-01-challenge) is used. This requires specifying the domain name(s) in the config _or_ ensuring that the server hostname is the intended domain name. 

However, the lib supports extensible ACME `challenge type` modules which a Gaia hub deployment system can use to easily add support for various DNS/DDNS services. Modules are already available for Google Cloud DNS, AWS (S3, Route53), Azure, CloudFlare, Digital Ocean, Namecheap, Godaddy, and more. 

Example of minimum config required to use (requires that server hostname is set to the intended domain name):
```json
{
  "enableHttps": "acme",
  "acmeConfig": {
    "email": "matt@example.com",
    "agreeTos": true
  }
}
```

Example of config where the domain name(s) are specified:
```json
{
  "enableHttps": "acme",
  "acmeConfig": {
    "email": "matt@example.com",
    "agreeTos": true,
    "approveDomains": ["gaia-testing.example.com", "gaia-testing2.example.com"]
  }
}
```

## Manual Testing

The ACME setup can be tested by setting up a Gaia hub on a VPS, pointing a domain to the server IP, and using the above config to enable `"enableHttps": "acme"`. The domain name can be specified in the hub config, or in the server's hostname config (e.g.: `hostname --fqdn` should show your domain name). 
I tested both methods using a Digital Ocean droplet and a namecheap subdomain A record. 

The TLS cert file setup can be tested by using the same resources from above, but with config set to `"enableHttps": "cert_files"`, and the `tlsCertConfig` options can be pointed to the cert PEM files that were generated by Lets Encrypt. 

## Automated Testing

The manual cert file config is covered in unit tests using self-signed localhost certs. 

The ACME setup is not covered in unit tests, except for the basic config value tests. The actual logic for using this lib is only a few lines of code -- essentially just:
```ts
// acmeClient.ts
import * as glx from 'greenlock-express'
export function createGlx(app: ExpressApp, acmeConfig: AcmeConfigInterface) {
  const opts: glx.GreenlockExpressOptions = { ... }
  return glx.create(opts)
}

// index.ts
const server = acme.createGlx(app, conf.acmeConfig)
server.listen(conf.port, conf.httpsPort, () => logHttpListen(), () => logHttpsListen())
```

The `greenlock-express` lib could be mocked to achieve code coverage, but it would be a useless test. This feature needs integration tests, which requires devops work for setting up a DNS API and likely a server, and it likely would not be possible to run automatically on commits due to rate limits. 

## Submission Checklist


jeffdomke:
@zone117x is this ready for code review?

hub/config-schema.json
```diff
@@ -2,6 +2,70 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "additionalProperties": false,
     "properties": {
+        "acmeConfig": {
+            "additionalProperties": false,
+            "description": "Options for Automatic Certificate Management Environment client. \nRequires `enableHttps` to be set to `acme`. \nSee https://www.npmjs.com/package/greenlock-express \nSee https://tools.ietf.org/html/rfc8555 \nSee https://github.com/ietf-wg-acme/acme",
+            "properties": {
+                "agreeTos": {
+                    "description": "Accept Let's Encrypt™ v2 Agreement. You must accept the ToS as the host which handles the certs. \nSee the subscriber agreement at https://letsencrypt.org/repository/",
```
jcnelson: Doesn't have to be done in this PR, but I'd prefer if we avoided unicode characters (e.g. `tm`) to the greatest extent possible.

hub/src/index.ts
```diff
@@ -11,8 +14,28 @@ validateConfigSchema(schemaFilePath, conf)
...
+  logger.warn('Setting up https server with ACME')
+  const server = acme.createGlx(app, conf.acmeConfig)
+  server.listen(conf.port, conf.httpsPort, () => logHttpListen(), () => logHttpsListen())
+} else if (conf.enableHttps === HttpsOption.cert_files) {
+  logger.warn('Setting up https server with provided cert files')
+  // Start Express app server with Node.js `https` and `http` modules.
+  tlsServer.createHttpsServer(app, conf.tlsCertConfig).listen(conf.httpsPort, () => logHttpsListen())
+  http.createServer(app).listen(conf.port, () => logHttpListen())
+} else {
+  // Start Express app server with only `http`.
```
jcnelson: Should there be a warning for this as well?

hub/src/server/tlsServer.ts
```diff
@@ -0,0 +1,97 @@
...
+function isTlsPfxCert(tlsConfig: TlsCertConfigInterface): tlsConfig is TlsPfxCert {
+  return !!(tlsConfig as TlsPfxCert).pfxFile
+}
+
+function isTlsPemCert(tlsConfig: TlsCertConfigInterface): tlsConfig is TlsPemCert {
+  return !!(tlsConfig as TlsPemCert).keyFile
+}
+
+export function loadPemCert(tlsConfig: TlsPemCert, opts: https.ServerOptions) {
+  if (tlsConfig.keyFile.startsWith(PEM_PREFIX)) {
```
jcnelson: Can we guarantee that `.keyFile` will be stripped of whitespace?

moxiegirl:
I am assigning this to me --- even though it is closed.  This will allow me to also track the need for docs.  It will appear in the docs pipeline.

hub/config-schema.json
zone117x: Fixed

hub/src/index.ts
zone117x: Fixed

hub/src/server/tlsServer.ts
zone117x: Fixed

Issue #267, opened 2019-09-17
Implement Gaia OCC / MVCC


zone117x:
Determine best approach to implement a Gaia file [MVCC](https://en.wikipedia.org/wiki/Multiversion_concurrency_control) system. 

Part of [Proposal: JSON Document Storage API with conflict prevention and resolution](https://github.com/blockstack/blockstack.js/issues/702)

### Option 1: Using a unique ID with an http header
Document files are stored in Gaia with an associated UUID that represents the version of that file. This would be stored in an HTTP header (e.g.: `ETag`, `x-bstk-mvcc-id`, etc). All driver implementations and their associated service APIs support using headers for these purposes. 

A blockstack.js `openDoc` function keeps track of this UUID. Once the document has been mutated, the UUID is sent along with the file upload request to Gaia. 

The Gaia server checks if the write-request-UUID matches the current file UUID. If it differs then it indicates that the app's in-memory document is in a conflicting state (uploading it would result in overwriting changes made to the document since it was originally fetched). The Gaia hub responds with a [HTTP `412 Precondition Failed`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412) response along with the latest UUID, and the content of the latest file. 

When the client receives the `Precondition Failed` response, it performs a document merge between the app’s document and the latest document returned by the Gaia server. The client then reattempts the write. 

Implementation note: the client would use a something like random backoff algorithm after repeated `Precondition Failed` responses in order to resolve race conditions with another client that may be attempting to resolve a conflict at the same time. 

On file writes, the Gaia hub would perform a file HEAD/stat request through the driver to retrieve the UUID from the response header. (Similar to how the revocation token system works).

This system leverages the file endpoint based mutex already implemented in Gaia to make this a fully atomic operation. 

An HTTP header for this purpose cannot easily be used to transmit the UUID due to the restrictive `CORs safelists` for [`response headers`](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header) and [`request headers`](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header). The client write request can pass the UUID to the hub using a URL query param. However, for the client to receive the UUID, both the Gaia HTTP server and the read endpoint server (e.g. an S3 bucket) need to whitelist whatever header using [`Access-Control-Expose-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers). 

[...]

PR #276, opened 2019-11-18
Concurrency control

reedrosenbluth:
## Description

This is the Gaia portion of the optimistic concurrency control changes I've been working on, referenced in this issue: blockstack/gaia#267

## Implementation

With this change, Gaia makes use of an optional etag (entity tag), included in the `if-match` request header, to validate that a file hasn't been changed externally since the client attempting to `write` to it last `read` it. Additionally, Gaia will now return a fresh `etag` in the response object returned on `write` operations.

This change helps prevent accidental overwrites and data loss, encouraging developers to create UIs to let users deal with these conflicts.

## Manual Testing

I've been testing this manually with a [custom Gaia hub](http://gaia.reedrosenbluth.com) running this code on digital ocean. I've been using a [small demo app](https://github.com/reedrosenbluth/bin) with [this version](https://github.com/blockstack/blockstack.js/pull/743) of blockstack.js

## Automated Testing

- [x] Modified existing unit tests to check for existence/correctness of etag

## Submission Checklist

- [ ] Submission contains tests that cover any and all new functionality or code changes.

- [ ] Submission documents any new features or endpoints, and describes how developers

### DIFF
--- a/hub/src/server/drivers/AzDriver.ts
+++ b/hub/src/server/drivers/AzDriver.ts
@@ -9,7 +9,7 @@ import {
   ListFileStatResult, DriverStatics, DriverModel, DriverModelTestMethods
 } from '../driverModel'
 import { Readable } from 'stream'
-import { BlobGetPropertiesHeaders } from '@azure/storage-blob/typings/src/generated/src/models'
+import { BlobGetPropertiesHeaders, BlobProperties } from '@azure/storage-blob/typings/src/generated/src/models'
 
 export interface AZ_CONFIG_TYPE {
   azCredentials: {
@@ -205,7 +205,7 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
       logger.debug(`Storing ${azBlob} in ${this.bucket}, URL: ${publicURL}`)
       return {
         publicURL,
-        etag: uploadResult.eTag
+        etag: uploadResult.eTag.replace(/^"|"$/g, '')
       }
     } catch (error) {
       logger.error(`failed to store ${azBlob} in ${this.bucket}: ${error}`)
@@ -270,14 +270,16 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
     }
   }
 
-  static parseFileStat(properties: BlobGetPropertiesHeaders) {
+  static parseFileStat(properties: BlobGetPropertiesHeaders | BlobProperties) {
     let lastModified: number | undefined
     if (properties.lastModified) {
       lastModified = dateToUnixTimeSeconds(properties.lastModified)
     }
+    let etag = (properties as BlobProperties).etag || (properties as BlobGetPropertiesHeaders).eTag
+    etag = etag.replace(/^"|"$/g, '')
     const result: StatResult = {
       exists: true,
-      etag: (properties.eTag || (properties as any).etag),
+      etag,
       contentLength: properties.contentLength,
       contentType: properties.contentType,
       lastModifiedDate: lastModified
--- a/hub/src/server/drivers/AzDriver.ts
+++ b/hub/src/server/drivers/AzDriver.ts
@@ -2,7 +2,7 @@
 
 import * as azure from '@azure/storage-blob'
 import { logger, dateToUnixTimeSeconds } from '../utils'
-import { BadPathError, InvalidInputError, DoesNotExist, ConflictError } from '../errors'
+import { BadPathError, InvalidInputError, DoesNotExist, ConflictError, PreconditionFailedError } from '../errors'
 import { 
   PerformWriteArgs, WriteResult, PerformDeleteArgs, PerformRenameArgs, PerformStatArgs,
   StatResult, PerformReadArgs, ReadResult, PerformListFilesArgs, ListFilesStatResult,
@@ -216,6 +216,9 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
       }
     } catch (error) {
       logger.error(`failed to store ${azBlob} in ${this.bucket}: ${error}`)
+      if (error.body && error.body.Code === 'ConditionNotMet') {
+        throw new PreconditionFailedError('The entity you are trying to create already exists')
+      }
       if (error.body && error.body.Code === 'InvalidBlockList') {
         throw new ConflictError('Likely failed due to concurrent PUTs to the same file')
       }
--- a/hub/src/server/drivers/AzDriver.ts
+++ b/hub/src/server/drivers/AzDriver.ts
@@ -194,7 +194,8 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
         blobCacheControl: this.cacheControl || undefined
       },
       modifiedAccessConditions: {
-        ifMatch: args.ifMatch
+        ifMatch: args.ifMatch,
+        ifNoneMatch: args.ifNoneMatch
       }
     }
 
--- a/hub/src/server/drivers/AzDriver.ts
+++ b/hub/src/server/drivers/AzDriver.ts
@@ -4,9 +4,9 @@ import * as azure from '@azure/storage-blob'
 import { logger, dateToUnixTimeSeconds } from '../utils'
 import { BadPathError, InvalidInputError, DoesNotExist, ConflictError } from '../errors'
 import { 
-  PerformWriteArgs, PerformDeleteArgs, PerformRenameArgs, PerformStatArgs, StatResult, 
-  PerformReadArgs, ReadResult, PerformListFilesArgs, ListFilesStatResult, ListFileStatResult,
-  DriverStatics, DriverModel, DriverModelTestMethods
+  PerformWriteArgs, WriteResult, PerformDeleteArgs, PerformRenameArgs, PerformStatArgs,
+  StatResult, PerformReadArgs, ReadResult, PerformListFilesArgs, ListFilesStatResult,
+  ListFileStatResult, DriverStatics, DriverModel, DriverModelTestMethods
 } from '../driverModel'
 import { Readable } from 'stream'
 import { BlobGetPropertiesHeaders } from '@azure/storage-blob/typings/src/generated/src/models'
@@ -163,7 +163,7 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
     }
   }
 
-  async performWrite(args: PerformWriteArgs): Promise<string> {
+  async performWrite(args: PerformWriteArgs): Promise<WriteResult> {
     // cancel write and return 402 if path is invalid
     if (!AzDriver.isPathValid(args.path)) {
       throw new BadPathError('Invalid Path')
@@ -187,7 +187,7 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
     const maxBuffers = 1
 
     try {
-      await azure.uploadStreamToBlockBlob(
+      const uploadResult = await azure.uploadStreamToBlockBlob(
         azure.Aborter.none, args.stream,
         blockBlobURL, bufferSize, maxBuffers, {
           blobHTTPHeaders: {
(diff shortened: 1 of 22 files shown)


hub/src/server/drivers/diskDriver.ts
```diff
@@ -274,8 +283,15 @@ class DiskDriver implements DriverModel {
     const contentTypeJsonStr = await fs.readFile(contentTypeFilePath, 'utf8')
     const contentType = JSON.parse(contentTypeJsonStr)['content-type']
     const lastModified = dateToUnixTimeSeconds(stat.mtime)
+
+    // create "weak" tag from last-modified time and file size
+    const mtime = stat.mtime.getTime().toString(16)
```
zone117x: I think this should just store a UUID as the etag within the corresponding json file metadata. We already eat all the costs from maintaining metadata state due to `content-type`. 
```js
// TODO: replace `contentTypeJsonStr` with something less specific 
const metadata = JSON.prase(contentTypeJsonStr)
const contentType = metadata['content-type']
const etag = metadata['etag']
```

hub/src/server/drivers/diskDriver.ts
```diff
@@ -243,7 +243,16 @@ class DiskDriver implements DriverModel {
       contentTypeFilePath, 
       JSON.stringify({ 'content-type': contentType }), { mode: 0o600 })
```
zone117x: (Related to the comment below about storing etags as uuids in the metadata file)
This could be something like 
```js
JSON.stringify({ 
  'content-type': contentType,
  'etag': crypto.randomBytes(32).toString('hex')
})

File: hub/src/server/server.ts [hunk 3]
```diff
@@ -213,6 +214,19 @@ export class HubServer {
...
+    const requestTag = requestHeaders['if-match']
+    if (requestTag) {
+      const responseTag = (await this.driver.performStat({
+        path: path,
+        storageTopLevel: address
+      })).etag
+
+      if (requestTag !== responseTag) {
+        console.log(requestTag)
+        throw new ValidationError(`The entity you are trying to store has been overwritten since your last read`)
```
zone117x: For this error type, I think we'd want something like a `PreconditionFailedError`, and the http layer would returns a 412 error with useful properties. 

Perhaps the expected `ETag` in the error response, so that future blockstack.js resolution callback code has the option of "wipe out existing changes"?

hub/src/server/server.ts
```diff
@@ -213,6 +214,21 @@ export class HubServer {
       }
     }
 
+    const requestTag = requestHeaders['if-match']
+    if (requestTag) {
```
zone117x: The DriveModel interface should have a new boolean property which indicates that the generic ETag validation should not be performed, and instead the driver will handle the check itself. Like `supportsETagConditionalUploads` or some better name. 

A couple use cases:
1. Improved latency (one less network request) for cloud storage APIs that support atomic conditional uploads using ETags. Azure supports this, so the PBC hub will have essentially no penalties when this goes live. Google Cloud API docs indicate it _might_ be supported (I can't tell without further testing). S3 protocol does not support it :(
2. Potential driver implementations that can perform a more efficient check for whatever reason. _Contrived example_: the `performStat` might be slow in order to retrieve the file size value, but really fast to check the etag.

reedrosenbluth: working on this!

hub/src/server/drivers/AzDriver.ts
```diff
@@ -187,7 +189,7 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
     const maxBuffers = 1
 
     try {
-      await azure.uploadStreamToBlockBlob(
+      const uploadResult = await azure.uploadStreamToBlockBlob(
         azure.Aborter.none, args.stream,
         blockBlobURL, bufferSize, maxBuffers, {
```
kantai: I believe Azure blob uploads support If-Match etag conditions directly, see: https://docs.microsoft.com/en-us/javascript/api/%40azure/storage-blob/blockblobuploadstreamoptions?view=azure-node-latest#conditions

We should implement this at least for Azure, since that'll have the most impact on the default user Gaia hub.

reedrosenbluth: Yes, I believe that's correct too. We included an optional `if-match` condition in the `DriverModel`s `performWrite` option type for this purpose. Was planning on implementing this in a followup PR, but perhaps it should be included in this one.

hub/src/server/server.ts
```diff
@@ -213,6 +214,23 @@ export class HubServer {
       }
     }
 
+    const requestTag = requestHeaders['if-match']
+    if (!this.driver.supportsETagMatching) {
+      if (requestTag) {
+        const freshTag = (await this.driver.performStat({
```
zone117x: Looking over the associated blockstack.js PR and thinking about the various scenarios during a `putFile` request -- I think we have to include structured data (JSON string) in the `if-match` header in order to handle a couple situations:

Blockstack.js has a client-side {filename,etag} map: 
```ts
const etags: { [filename: string]: string } = {}
``` 

If `putFile` is performed with no corresponding etag, then it could mean at least a couple scenarios:

1. The dev's intention is to create a new file and no existing file expected to exist. However, there may be a concurrency bug _(unanticipated by the dev, as with most concurrency bugs)_ where the new file was already created in another session. I think with the current code, the second "new file" request will have an empty `if-match`, causing this etag check to be skipped and clobbering the original file. 
   * This could be solved if the `if-match` content was structured, for example `JSON.parse(headers['if-match']).etag === null` could indicate that a check should be performed. That check could look something like `(await driver.performStat(...)).exists === false`. 

2. A use-case where file overwrites are intentional, where the local `etag` file map is often outdated -- for example high frequency intentional overwrites coming from multiple sessions. I think we've decided overwrites need to be explicitly specified by devs, with an API that clearly indicates the consequences, e.g. `putFile(..., { dangerousOverwrite: true })`. With this current etag check, blockstack.js would have to implement this by omitting the `if-match` value. 
I think the Gaia http spec should also be explicit about overwrites, and "overwrite by omitting `if-match`" is feels like overloading of legacy behavior. 
   * I think structured `if-match` data could also solve this. For example `JSON.parse(headers['if-match']).overwrite === true` could be checked.

kantai: In general, I'm in favor of trying to preserve the standard behavior and syntax of the `If-Match` header -- i.e., accepting just an ETAG.

The standard way to deal with the 'new file' case is to use the `If-None-Match` header, as `If-None-Match: *`

Would `blockstack.js` be able to differentiate the intention of creating a new file from the intention of updating a file?

zone117x: `If-None-Match` header is a good idea. I think that could solve the 'new file' scenario. I've generally wanted to avoid new headers due to the additional CORs config overhead, but makes sense in this case. 

> Would blockstack.js be able to differentiate the intention of creating a new file from the intention of updating a file?

At the moment, blockstack.js only knows of the missing local etag map entry, and blockstack.js can only assume the dev's intention is create a new file. This assumption will not always be true. Blockstack.js could implement a `createFile` function (or equivalent option in `putFile`) for devs to specify their intention -- then `putFile` can just throw immediately, rather than assume, attempt upload, and throw. 

---

The remaining problem is how to specify an explicit overwrite. I think this could be performed by specifying `If-Match: *`?

So the possible blockstack.js-vNext `putFile` requests must be one of:
* `If-Match: <etag-value>`: typical file update -- the etag map entry exists.
* `If-None-Match: *` as new file upload -- assumed from missing etag map entry.
* `If-Match: *` as file overwrite -- dev explicitly specified overwrite flag with  `putFile`.

hub/src/server/server.ts
```diff
@@ -213,6 +214,42 @@ export class HubServer {
       }
     }
 
+    const ifMatchTag = requestHeaders['if-match']
+    const ifNoneMatchTag = requestHeaders['if-none-match']
+
+    // only one match-tag header should be set
+    if (ifMatchTag && ifNoneMatchTag) {
```
zone117x: Nice defense programming here 👍

hub/src/server/server.ts
```diff
@@ -213,6 +215,42 @@ export class HubServer {
...
+        // only proceed with writing file if the file does not already exist
+        const statResult = await this.driver.performStat({
+          path: path,
+          storageTopLevel: address
+        })
+
+        if (statResult.exists) {
+          throw new PreconditionFailedError('The entity you are trying to create already exists')
+        }
+      } else if (ifNoneMatchTag && ifNoneMatchTag !== '*') {
```
zone117x: I think this condition is supposed to be outside of the `if (!this.driver.supportsETagMatching)` condition scope. As in, next to the `if (ifMatchTag && ifNoneMatchTag)` check above.

hub/src/server/server.ts
```diff
@@ -213,6 +215,42 @@ export class HubServer {
...
+    const ifNoneMatchTag = requestHeaders['if-none-match']
+
+    // only one match-tag header should be set
+    if (ifMatchTag && ifNoneMatchTag) {
+      throw new PreconditionFailedError('Request should not contain both if-match and if-none-match headers')
+    }
+
+    if (!this.driver.supportsETagMatching) {
+      // allow overwrites if tag is wildcard 
+      if (ifMatchTag && ifMatchTag !== '*') {
```
zone117x: Should we have an explicit `ifMatchTag === '*' ... then overwrite` condition? 
Looks like the overwrite scenario may fall through to what I'm considering "legacy" behavior, but IMO we should have an explicit check

hub/src/server/server.ts
reedrosenbluth: Why? Don't we only want to check for the existence of a file if the driver can't do it for us?

hub/src/server/server.ts
reedrosenbluth: How would you implement this? What would be in the body of the `ifMatchTag === '*'` conditional? 

I thought about doing this, but wasn't sure how to do it without repeating any code. Since the (over)write code happens below, I figured I'd simply let wildcard tags pass through via the conditional

zone117x: Hmm, you're right, looks like the body would be a noop, or oddly duplicated write code..
Perhaps the condition would be more like 
```js
if (!ifMatchTag && !ifNoneMatchTag) { 
  log('deprecated/legacy write request')
}
```

This is in anticipation of a future where that `log` is replaced with an error, ensuring future Gaia client implementations don't end up with the same concurrency bugs as (current) blockstack.js. 

I don't feel that strongly about this -- perhaps there are good arguments for why this condition should be indefinitely allowed? @kantai @jcnelson

hub/src/server/server.ts
zone117x: I think without the check, this would result in some gaia hubs accepting `ifNoneMatchTag !== '*'`, while some hubs would not, depending on the driver implementation.

hub/src/server/server.ts
kantai: The way the condition is checked now works for me. I'd be against adding that log entry for now, since it would generate a _lot_ of entries. Plus, this could be logged on an HTTP front-end.

hub/src/server/drivers/AzDriver.ts
```diff
@@ -186,16 +188,32 @@ class AzDriver implements DriverModel, DriverModelTestMethods {
...
+      // Return success and url to user
+      const readURL = this.getReadURLPrefix()
+      const publicURL = `${readURL}${azBlob}`
+      logger.debug(`Storing ${azBlob} in ${this.bucket}, URL: ${publicURL}`)
+      return {
+        publicURL,
+        etag: uploadResult.eTag.replace(/^"|"$/g, '')
+      }
     } catch (error) {
       logger.error(`failed to store ${azBlob} in ${this.bucket}: ${error}`)
```
zone117x: Should this check for Azure's conditional failure error and throw the same error as the generic version?
e.g.
```js
if (isAzureConditionalFailureResponse(error)) {
  throw new PreconditionFailedError(...)
}
```

Otherwise I think this would result in gaia returning a 500?

PR #281, opened 2019-12-03
Update unfair table

DaWe35:
## Goal of PR

Correcting information in the comparing table. (I'll be happy to remove the whole table because it compares apples to pears. Pointless, more below.)

## Changes

> Data can be viewed in a normal Web browser

Sorry guys, but that's a stupid question. Every decentralized storage solution can be viewable in a browser with a CDN (or middle server). Everything that is really decentralized needs a full node. Because of that, it's not viewable in a browser, without a full node. Gaia will be decentralized when I can add multiple storages to it, but that's optional.

> Data can be deleted

On Sia, files can be deleted, I think this is just a simple mistake.

> Data lookups have predictable performance

I'm watching 1080p films from a mounted Sia drive. That's improved a lot and will be improved more.

> Supports multiple backends natively

Ahh, here we go again. What does this question mean? On Sia, every file stored on 30 geographically distributed hosts. I think it is multiple. You can increase/decrease the redundancy also.
I removed this line, because every competitor is a backend, while Gaia is the client. You can use Sia with Gaia (with a mounted s3 drive), that's a stupid comparison. Apple is not comparable with pear.

> Data is read/write

What does this question mean?

I also added some new rows.

Note that I have strong experience with Sia, but I don't know much about Dat and SSB (that's a social media, not storage, why is it on the list? Apple and pear again.) IPFS is a protocol, FileCoin is the storage (yeah, not working yet, but again, that's not comparable).

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,25 +585,15 @@ Currently, the service will only allow up to 20 write requests per second and a 
 file size of 5MB. However, these limitations are only for our service, if you deploy
 your own Gaia hub, these limitations are not necessary.
 
-# Project Comparison
+# Features
 
-Here's how Gaia stacks up against other decentralized storage systems.  Features
-that are common to all storage systems are omitted for brevity.
-
-| **Features**                | Gaia | [Sia](https://sia.tech/) | [Storj](https://storj.io/) | [IPFS](https://ipfs.io) | [DAT](https://datproject.org) | [SSB](https://www.scuttlebutt.nz/) |
-|-----------------------------|------|--------------------------|----------------------------|-------------------------|-------------------------------|------------------------------------|
-| User controls where data is hosted           | X |   |   |   |   |   |
-| Data stored redundantly                      | optional | X | X | optional | optional | optional |
-| Data is read/write                           | X |   |   |   | X | X |
-| Data can be deleted                          | X | X |   |   | X | X |
-| Data can be listed                           | X | X | X |   | X | X |
-| Deleted data space is reclaimed              | X | X | X | X |   |   | 
-| Data lookups have predictable performance    | X | X | X |   |   |   |
-| Writes permission can be delegated           | X |   |   |   |   |   |
-| Listing permission can be delegated          | X |   |   |   |   |   |
-| Data is globally addressable                 | X | X | X | X | X |   | 
-| Needs a cryptocurrency to work               |   | X | X |   |   |   |
-| Needs FIAT currency to work                  | X |   |   |   |   |   |
-| Data is content-addressed                    |   | X | X | X | X | X |
-| Data stored geographically distributed       | optional | X | X | X | X | X |
-| Pricing (per TB per month)                   | Various | ~$2 | $10 | free, files can be lost | free, files can be lost |  |
+- User controls where data is hosted
+- Data can be viewed in a normal Web browser
+- Data can be stored redundantly (depends on the storage you choose)
+- Data can be listed and deleted
+- Deleted data space is reclaimed
+- Data lookups have as predictable performance as your storage
+- Writes and listing permission can be delegated
+- Data is globally addressable
+- Useable with free/trial storage packages
+- Noo needs cryptocurrency to work, just storage
--- a/README.md
+++ b/README.md
@@ -593,15 +593,17 @@ that are common to all storage systems are omitted for brevity.
 | **Features**                | Gaia | [Sia](https://sia.tech/) | [Storj](https://storj.io/) | [IPFS](https://ipfs.io) | [DAT](https://datproject.org) | [SSB](https://www.scuttlebutt.nz/) |
 |-----------------------------|------|--------------------------|----------------------------|-------------------------|-------------------------------|------------------------------------|
 | User controls where data is hosted           | X |   |   |   |   |   |
-| Data can be viewed in a normal Web browser   | X |   |   | X |   |   |
+| Data stored redundantly                      | optional | X | X | optional | optional | optional |
 | Data is read/write                           | X |   |   |   | X | X |
-| Data can be deleted                          | X |   |   |   | X | X |
+| Data can be deleted                          | X | X |   |   | X | X |
 | Data can be listed                           | X | X | X |   | X | X |
 | Deleted data space is reclaimed              | X | X | X | X |   |   | 
-| Data lookups have predictable performance    | X |   | X |   |   |   |
+| Data lookups have predictable performance    | X | X | X |   |   |   |
 | Writes permission can be delegated           | X |   |   |   |   |   |
 | Listing permission can be delegated          | X |   |   |   |   |   |
-| Supports multiple backends natively          | X |   | X |   |   |   |
 | Data is globally addressable                 | X | X | X | X | X |   | 
 | Needs a cryptocurrency to work               |   | X | X |   |   |   |
+| Needs FIAT currency to work                  | X |   |   |   |   |   |
 | Data is content-addressed                    |   | X | X | X | X | X |
+| Data stored geographically distributed       | optional | X | X | X | X | X |
+| Pricing (per TB per month)                   | Various | ~$2 | $10 | free, files can be lost | free, files can be lost |  |


README.md
```diff
@@ -593,15 +593,17 @@ that are common to all storage systems are omitted for brevity.
 | **Features**                | Gaia | [Sia](https://sia.tech/) | [Storj](https://storj.io/) | [IPFS](https://ipfs.io) | [DAT](https://datproject.org) | [SSB](https://www.scuttlebutt.nz/) |
 |-----------------------------|------|--------------------------|----------------------------|-------------------------|-------------------------------|------------------------------------|
 | User controls where data is hosted           | X |   |   |   |   |   |
-| Data can be viewed in a normal Web browser   | X |   |   | X |   |   |
+| Data stored redundantly                      | optional | X | X | optional | optional | optional |
```
jcnelson: In each system Gaia supports, the data is _definitely_ stored redundantly within the storage provider.  This cannot be said about any of the others.  Even Sia doesn't guarantee that all N erasure codes aren't simply all placed on one host.

README.md
```diff
@@ -593,15 +593,17 @@ that are common to all storage systems are omitted for brevity.
...
 | Data is read/write                           | X |   |   |   | X | X |
-| Data can be deleted                          | X |   |   |   | X | X |
+| Data can be deleted                          | X | X |   |   | X | X |
```
jcnelson: In my mind, "deleted" means "once the request is acknowledged, a subsequent read will always fail."  Is this true for Sia?  Or does Sia just garbage-collect erasure codes in the background?  If it's the latter, then I don't consider this to be "deleted."

README.md
```diff
@@ -593,15 +593,17 @@ that are common to all storage systems are omitted for brevity.
...
 | Data can be listed                           | X | X | X |   | X | X |
 | Deleted data space is reclaimed              | X | X | X | X |   |   | 
-| Data lookups have predictable performance    | X |   | X |   |   |   |
+| Data lookups have predictable performance    | X | X | X |   |   |   |
```
jcnelson: I don't believe this is true in general, your (single data point of) experiences not withstanding.  Does Sia offer SLAs?  Is it even possible in the system design of Sia to _offer_ SLAs?

README.md
```diff
@@ -593,15 +593,17 @@ that are common to all storage systems are omitted for brevity.
...
 | Writes permission can be delegated           | X |   |   |   |   |   |
 | Listing permission can be delegated          | X |   |   |   |   |   |
-| Supports multiple backends natively          | X |   | X |   |   |   |
 | Data is globally addressable                 | X | X | X | X | X |   | 
 | Needs a cryptocurrency to work               |   | X | X |   |   |   |
+| Needs FIAT currency to work                  | X |   |   |   |   |   |
```
jcnelson: NACKing this change.  It doesn't tell us anything about the behavior of the storage medium.  Moreover, for every storage provider and every form of currency, there exists some sequence of actions you can take that will enable you to pay for the storage of data in that storage provider.

README.md
```diff
@@ -593,15 +593,17 @@ that are common to all storage systems are omitted for brevity.
...
 | Data is content-addressed                    |   | X | X | X | X | X |
+| Data stored geographically distributed       | optional | X | X | X | X | X |
```
jcnelson: This is definitely _not_ guaranteed by IPFS -- your data is only as distributed as your pinners (and for the long tail of content, there is one pinner: you).  I don't think it's guaranteed by Sia either, unless you can demonstrate that the erasure codes actually _do_ get geo-replicated (which is basically impossible to prove -- miners can simply run many VPN endpoints in different countries, for example).  I think Storj will give you an interface for doing this, but I don't know how they enforce storage nodes to conform to it (same problem).  SSB and DAT are questionable as well -- it depends on how many replicas you can convince other people to create and persist.

README.md
```diff
@@ -593,15 +593,17 @@ that are common to all storage systems are omitted for brevity.
...
+| Pricing (per TB per month)                   | Various | ~$2 | $10 | free, files can be lost | free, files can be lost |  |
```
jcnelson: Once again, this isn't relevant to the behavior of the storage medium.  Also, it would be "various" for everything -- the price of storage fluctuates over time, as do the prices of cryptocurrencies.

README.md
DaWe35: Hmm, that's true, but with S3, Google Drive, etc there is always a single point of failure. With Sia, the failure is expected, and after 25% of the hosts disappeared, redundancy automatically boosted. Maybe it's geographically redundant (you don't know, they just said that), one company is responsible for it.
I asked the Sia core devs about that. Theoretically, that could happen that there are 30 hosts in one hand, but that would be expensive, and Sia also does a lot to explore host farms. Each host needs to be geographically distributed, and run minimum 2-3 months to get any renter. These hosts also need to lock some collateral, and when they go offline, it'll be lost.
That's interesting, so I can't decide who's right. Pls help @tbenz9 :)

README.md
DaWe35: >Using the Sia blockchain, renters form file contracts with hosts. These contracts set pricing, uptime commitments, and other aspects of the relationship between the renters and the hosts.
File contracts are a type of smart contract. They allow us to create cryptographic service level agreements (SLAs) that are stored on the Sia blockchain.
Since file contracts are automatically enforced by the network, Sia has no need for intermediaries or trusted third parties.

Source: https://sia.tech/technology

If the SLA fails, the host will lose the locked collateral. And of course renters prefer hosts with moer collateral.

jcnelson: Thanks for the link.

The reason I asked the second question is because while it's all well and good for Sia to _claim_ to offer SLAs, I don't think the design of the system itself makes this possible to do beyond a binary outcome.  The SLA Sia seems to offer, from what I could gather from their website, is that either

* the file is available if you wait long enough, or
* the file is not available; slash the miner

This is not what I would call "predictable performance," as understood by a reasonable user.  "Predictable performance" here means "I can request the file from any computer in the world and at any time, and it will almost always download in the same reasonable amount of time."  For example, Dropbox has predictable performance -- you can request to download a 1GB file from anywhere in the world at any time from any device, and there's a good chance (e.g. 99% or higher) it will (1) be available and (2) download relatively quickly -- as fast as your Internet connection allows.  As a counterexample, BitTorrent does not have predictable performance -- the download times and availability of a file is highly variable, and depends on the number of seeders that are online.  Unless the file is well-seeded, you won't saturate your network bandwidth either.

Under this understanding, does Sia provide predictable performance similar to e.g. Dropbox?  Or is it more like BitTorrent?  From what I have been able to find online, it looks more like BitTorrent.

I'd be willing to alter Gaia's `X` here to say something like `depends on the storage back-end`, because the read performance of Gaia is wholly dependent on the storage provider (which itself can offer less-than-predictable performance, if the user so chooses).

MeijeSibbel: No blockchain protocols can offer SLA's. Simple fact is, SLA's are more than just performance and availability guarantees; Support, penalties, customization, security, compliance and accountability are all part of SLA's.

Regarding performance, you guys are looking at this the wrong way. Sia is a storage marketplace with a vast number of different storage nodes with different performance metrics. It is not a storage service provider, that is layer two, e.g. a service built on Sia. 

If you are layer two and you have your own storage node selection layer built on Sia with benchmarking etc. you **will** get predictable performance. If you use the siad or the UI offered by Sia, not so much.

So if we talk strictly the API's offered by Sia @jcnelson is correct.

README.md
MeijeSibbel: Agreed on IPFS. I know Storj is already/is going to be vetting storage nodes of a certain size with identity-tied physical contracts they have to sign. With Sia this is going to be a layer two job again, not something that happens on the protocol lever. That said, it is easy to find if someone is running a VPN; Just look at application latency and latency to first byte. This is fairly easy to filter out.

README.md
MeijeSibbel: Agreed. Not even to mention any performance tiers for storage etc.

README.md
MeijeSibbel: Correct. At this time it is garbage-collection, if the contract is terminated, deleted data will remain on the storage node until the contract expires. "Delete" function is being worked on by the Sia team. Estimate is a few months until deployment. Theory behind this is complex but once delete is implemented it will be deleted node-side.

README.md
MeijeSibbel: What does 'guarantee' mean and in what way? Firstly redundancy can be set manually with Sia, as well as selecting storage nodes. Each Sia host has a unique host-ID validated on the blockchain with spent and locked collateral as proof. You can't have two storage hosts with the same ID. Just from that it can be considered as redundancy. If the redundant sector is stored in the same datacenter or not, it doesn't matter, it is stored on two separa node instances. The big three guarantee redundancy on paper, but i'd much rather trust a cryptographic proof that a different storage node has a copy of my data.

MeijeSibbel:
Just a quick final comment here. There is a difference between storage protocols/networks and storage service providers. Comparing for example Sia to AWS S3 is like comparing apples to oranges, they are no the same thing. AWS S3 is a service, Sia is a protocol layer for a storage marketplace. To be honest i would just take Sia out of this list.

README.md
jcnelson: > Firstly redundancy can be set manually with Sia, as well as selecting storage nodes. Each Sia host has a unique host-ID validated on the blockchain with spent and locked collateral as proof.

I definitely believe that Sia will let you choose the number of erasure codes you want to create, and choose the number that will be required to reconstruct the file.  I could be convinced that there is a way for you to choose which host-IDs end up with a copy of your erasure codes.  However, I do not believe that Sia gives you the ability to place those erasure codes on the physical hosts of your choosing, since the hosts _as Sia sees them_ are in no way guaranteed to be the hosts _as they actually are_.  For example, it is trivial to run many Sia nodes as VMs on one physical server, and submit convincing proofs to the network that each VM is a wholly-separate host running in a unique geographic location (via the VPN trick I mentioned elsewhere).  If each VM stored one of my file's erasure codes, I would _wrongfully believe_ that my file is distributed across many different physical locations, when it is in fact stored in one location (on one disk, even).

This is why I don't think Sia deserves the `X` for data being stored redundantly.  Neither the _protocol_ nor the _widely-used instance_ of Sia guarantee a reasonable kind of "redundancy".  After all, the _whole point_ of hosting something redundantly is to ensure that if the physical hardware fails, the data is still available ;)

README.md
jcnelson: Ah, good to know.  Thank you for the clarification!

README.md
jcnelson: > Regarding performance, you guys are looking at this the wrong way. Sia is a storage marketplace with a vast number of different storage nodes with different performance metrics. It is not a storage service provider, that is layer two, e.g. a service built on Sia.

To avoid future confusion, is there a different name for Sia-the-widely-used-blockchain-network than simply "Sia?"  Up until now, I thought the word "Sia" could be used to refer to either the protocol and the widely-used blockchain network.  This is the case for other p2p protocols, for example -- BitTorrent is the protocol, but The Pirate Bay is a particular BitTorrent network (with a particular collection of torrents and peers).

But to be clear, I think this table is correct for Sia regardless of whether you interpret it to mean the protocol or the blockchain network.

> If you are layer two and you have your own storage node selection layer built on Sia with benchmarking etc. you will get predictable performance. If you use the siad or the UI offered by Sia, not so much.

Are there such widely-used systems?  Happy to add columns for them, based on their particular feature sets.

README.md
jcnelson: > That said, it is easy to find if someone is running a VPN; Just look at application latency and latency to first byte. This is fairly easy to filter out.

I agree that there are heuristics that you can use to identify _likely_ VPN endpoints.  However, "likely" is un-quantifiable here.  No p2p storage system that I'm aware of (or any sort of wide-area telemetry protocol, for that matter) can guarantee with any certainty that a particular IP address corresponds to a particular server on the Internet -- not even to within a measurable error bound.

README.md
MeijeSibbel: >since the hosts as Sia sees them are in no way guaranteed to be the hosts as they actually are

This is not a valid argument as it is true for any storage system, including Gaia, including AWS, including Azure, Storj and GCP etc. With enough work anyone can be tricked. Furthermore i believe since centralized solutions are a lot less transparent, it is significantly easier to fool the customer. At the end of the day what would be the financial incentive to trick someone?, none, neither for centralized nor decentralized solutions.

You're trying to find counter arguments on a theoretical basis that can happen in any global storage network.

README.md
MeijeSibbel: I disagree. Comparing services as Gaia that use centralized storage providers to blockchain protocols like Sia does not make any sense. There are a few services that have built AWS S3 API's on Sia like https://filebase.com/ which do offer predictable performance.

README.md
MeijeSibbel: Again, as i mentioned above already, this is true for every storage provider including Gaia. Even AWS can trick you into believing the data is stored on a server it is not. It is actually a lot easier for AWS to trick you than Sia since there is zero proof or transparency from AWS on which storage node the data is stored.

README.md
jcnelson: They're not theoretical.  I can, with a court order, physically inspect these datacenters and determine where my data replicas are stored.  I can also sue AWS, Azure, GCP, etc., for lying to me or even get them thrown in jail for _not_ storing my data redundantly.  That these systems have a notion of legal liability enforced through legally-binding contracts is _ENORMOUSLY_ beneficial, since the threat of legal retribution for deception is, for all intents and purposes, sufficient to incentivize these storage providers to behave honestly even if I don't go out and inspect them (and even be forthright if they cannot keep up their end of the contract).

Sia does not have any of this.  In Sia and other decentralized storage networks, there's no downside to successfully deceiving users for profit if they can also deceive the (not-legally-binding) smart contract.  There's no recourse for fraud if the person running the nodes you thought were redundant were, in fact, not redundant.

To conclude this thread, we have agreed that Sia does not, at the protocol-level, make any guarantees of redundancy.  Moreover, there is not a reliable incentive mechanism outside of the protocol (e.g. legal contracts) that compels the Sia network to store data redundantly in spite of this.  Therefore, Sia does not store data redundantly.

README.md
jcnelson: Summarizing my comment above, I can sue AWS for tricking me.  I cannot sue anyone in the Sia network.  Moreover, Sia miners can profit by defrauding me without recourse.

README.md
tbenz9: Sounds to me like we have fundamental differences on what qualifies as "stored redundantly". 
 I'm not sure we're going to be able to resolve this one and ultimately @jcnelson controls the repo so he can decide what he deems appropriate and fair.

With that said, I've inspected the Sia codebase and I truly believe Sia data is "stored redundantly".  I also understand your reasoning, and while I can understand your hesitation, I don't believe it's valid.

DaWe35:
> 
> 
> Just a quick final comment here. There is a difference between storage protocols/networks and storage service providers. Comparing, for example, Sia to AWS S3 is like comparing apples to oranges, they are not the same thing. AWS S3 is a service, Sia is a protocol layer for a storage marketplace. To be honest i would just take Sia out of this list.

Yes, I agree, but I also would take out Storj, IPFS, DAT, and SSB:

- IPSF is a p2p file-sharing protocol.
- DAT is a distributed data synchronization protocol.
- SSB is a communication protocol and a self-hosted social media ecosystem.
- Storj and Sia are Decentralized cloud storage platforms.
- Gaia is a storage system (layer?)

You said that Gaia is redundant. What if I use Gaia with an S3 bucket build on Sia? ;)

This table does not make sense, for me looks like marketing bullshit. Don't get me wrong: I love what you do, and I'm planning to build dapps on Blockstack with Gaia, but for a newbie, it will be difficult to understand what Gaia really is.

README.md
jcnelson: What guarantees does Filebase make that Sia does not?  Can I measure them or audit them?  If Filebase is a distinct storage medium from Sia that makes different guarantees about e.g. redundancy, then it would deserve its own column.

README.md
jcnelson: Question for y'all -- if you copy a file 10 times and store all copies on your computer with the original, is that file stored redundantly?  If you were in the business of selling storage capacity and you told your customers that their data was stored redundantly this way, do you think your customers would pay you for the redundancy?  If you think the customer would say "no," then we have the same understanding of redundancy:  a file is only stored redundantly only if its data is stored on _different computers_.

As discussed, Sia does not make any guarantees about the placement of erasure codes on different computers (and it is entirely possible -- even profitable -- for Sia miners to "fake" being multiple miners).  Now, maybe systems like Filebase _can_ make this guarantee by e.g. running their own Sia miners that will actually store your erasure codes redundantly if you pay Firebase for the service.  But as also argued elsewhere, this makes Firebase a _fundamentally different storage system_ than Sia -- it offers _fundamentally different guarantees_ about the way your data is stored -- and would thus deserve its own column.

DaWe35: "Data redundancy, database systems which have a field that is repeated in two or more tables" - so Wikipedia says yes, and that was taught to me at university also, so I think yes, this is redundancy, even all the data stored in one computer. (if you send a file, and then you send the hash of it, that's also a redundant file transfer).
I think what we are talking about is distributed redundancy, and I think there is a way to get there with Sia. For example, if there is no 70% host superiority, you just need to pick random hosts, and your data stored redundantly because 10 of your 30 files are not on the same computer.

jcnelson:
The distinction between the protocol and the market-dominant system that implements the protocol is a distinction without a difference.  For example, when you say "store to S3" by itself, it is generally understood to mean "store to the S3 service offered by AWS" and not the S3 protocol.

>  it will be difficult to understand what Gaia really is.

Gaia is an authenticated, multiplexing storage proxy that operates on the write path, and optionally on the read path.  You POST data to a Gaia hub, and if you are authorized, it forwards your data to the storage provider(s) it is configured to talk to.  As a user, you can choose which Gaia hub a Blockstack app will send your writes to.

In particular, you can choose a Gaia hub (or run your own hub) that is able to guarantee data redundancy through its storage provider.  Moreover, the most widely-used Gaia hub (the one provided by Blockstack PBC) uses Azure as its storage provider, which does guarantee data redundancy.  This is not true for Sia-the-protocol, nor is it true for Sia-the-widely-used-blockchain.

DaWe35:
I wanted to suggest that as a user I can choose where Gaia stores my data. But what is I choose Sia (which you don't think is redundant)? Gaia can't guarantee redundancy itself, because I can use it non-redundant. And Sia also can't guarantee, and Bitcoin also can't guarantee that it is decentralized, because 99.9999...% is not 100%.
> In particular, you can choose a Gaia hub (or run your own hub) that is able to guarantee data redundancy through its storage provider.
How Gaia guarantees that my hub is redundant?

README.md
jcnelson: Because we are talking about a distributed storage system, it is reasonable to expect readers of this README to think "redundancy" means "distributed redundancy."  It is also reasonable to expect readers to think that "redundancy" means "the storage system will make sure my data is replicated to multiple physical computers, 100% of the time, by design" instead of best-effort (which Sia does).

That said, I'm totally fine with changing Gaia's `X` to `depends on the driver` for redundancy, because you can use a Gaia hub that does _not_ use a redundant storage system (even though you'd have to go out of your way to do this).  But Sia does not deserve the `X`.

jcnelson:
> I wanted to suggest that as a user I can choose where Gaia stores my data. But what is I choose Sia (which you don't think is redundant)?

As mentioned elsewhere, I'd be fine with accepting a PR that changed Gaia's `X` for redundancy to `depends on the driver` (or words to that effect).

DaWe35:
@jcnelson Thank you for your patience, maybe the first commit wasn't too useful, but I still think that comparing apples to pears does not make sense while Gaia is a delicious orange.
I really love user-friendly decentralized solutions, I think Gaia is very unique, there is no need to compare it to completely different projects.
What do you think?

zone117x:
Closing this issue due to lack of activity. Please feel free to comment if this should be re-opened.

PR #284, opened 2019-12-13
Feature/fix express cors

zone117x:
Fix Express server CORS config on `hub` and `reader` so that they work as expected when hosted without a http proxy (e.g. running on port 80/443 without nginx).


moxiegirl:
@reedrosenbluth @zone117x I need an updated JSON for the cors config.  Do we have one we are generating?

PR #365, opened 2022-02-25
Patch instructions and start-stop script

criadoperez:
Organized installation instructions.


deploy/README.md
```diff
@@ -1,72 +1,81 @@
...
+These driver may require you to provide additional credentials for performing writes to the backends. See `config.sample.json` for fields for those credentials. In some cases, the driver can use a system configured credential for the backend (e.g., if you are logged into your AWS CLI account, and run the hub from that environment, it won't need to read credentials from your `config.json`).
+
+*Note:* The disk driver requires a *nix like filesystem interface, and will not work correctly when trying to run in a Windows environment.
+
+### Note on SSL
+We *strongly* recommend that you deploy your Gaia hub with SSL enabled. Otherwise, the tokens used to authenticate with the Gaia hub may be stolen by attackers, which could allow them to execute writes on your behalf.  
+Configuration options are available to run the hub with an `https` Node.js server.  
+Otherwise, a reverse proxy web server such as nginx or Apache can be used.  
+
+### Require Correct Hub URL
```
jcnelson, line 44: I'm surprised that this is even an option.  I don't remember why this isn't on by default, but maybe it should be (cc @wileyj)

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+TASK=$1
+WHICH=$(which docker)
```
jcnelson, line 4: `which` isn't available on all *nix systems these days.  Recommend replacing with `command -v`, which is POSIX-ly correct.

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+TASK=$1
```
jcnelson, line 3: It's best to quote arguments (`TASK="$1"`).  Can you run `shellcheck` on this script and correct any warnings it reports?

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
```
jcnelson, line 2: Should the script include some extra strictness checks be turned on, such as `set -uo pipefail`?

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
...
+	echo " To stop the GAIA Hub type: $0 stop."
+	echo " To check if GAIA Hub is running type: $0 status."
+	echo " Simply typing $0 displays this help message."
+	echo
+}
+
+#Checks files I need exist
+check_files_exist() {
+	# If a file I need is missing, inform the user.
+	if ! test -f "$FILE1"; then
```
jcnelson, line 24: Strictly speaking, you don't need `test` here.  `if ! [ -f "$FILE1" ]` works just as well.

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
...
+	echo " To check if GAIA Hub is running type: $0 status."
+	echo " Simply typing $0 displays this help message."
+	echo
+}
+
+#Checks files I need exist
+check_files_exist() {
+	# If a file I need is missing, inform the user.
+	if ! test -f "$FILE1"; then 
+		echo "Missing $FILE1. Did you delete it?"
```
jcnelson, line 25: If these are error messages, then can this be directed to stderr (i.e. `echo >&2`)?

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+TASK=$1
+WHICH=$(which docker)
+SCRIPTPATH=$(pwd -P)
+FILE1="docker-compose-base.yaml"
```
jcnelson, line 6: Suggest using more descriptive variable names.  I don't know the difference between `FILE1` and `FILE2` without having to consult this code snippet.

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
...
+	if ! test -f "$FILE2"; then
+		echo "Missing $FILE2. Did you delete it?"
+		return 0
+	fi
+	if ! test -f "$FILE3"; then
+		echo "Missing $FILE3. Looks like you forgot to create one."
+		return 0
+	fi
+	# If all files I need exist, then continue
+	return 1
```
jcnelson, line 37: In shell scripts, returning non-zero is conventionally considered an error.  The semantics of shell function return values are the same as those for `exit(2)`, since they are often executed within subshells or even child processes.  Returning 0 conventionally indicates success.  Can you update this?

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
...
+	# docker not running
+	return 1
+}
+
+gh_status() {
+	if check_containers; then
+		echo "GAIA HUB running."
+		return 1
+	fi
+	if ! check_containers; then
```
jcnelson, line 56: Shouldn't this just be an `else` instead of an additional `if` statement?

deploy/docker/gaiahub.sh
criadoperez: I had it generic on purpose, but ok. I changed it to a more descriptive name if that is preferred.

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+TASK=$1
+WHICH=$(which docker)
+SCRIPTPATH=$(pwd -P)
+FILE1="docker-compose-base.yaml"
+FILE2="docker-compose-disk.yaml"
+FILE3="disk.env"
+
```
jcnelson, line 9: Once thing I usually do at the start of any shell script that uses programs not found in `/bin` or `/sbin` is I check that they're installed. Something like this:

```
exit_error() {
   printf "%s" "$1" >&2
   exit 1
}

for cmd in ncat grep tr dd sed cut date sqlite3 awk xxd; do
   command -v "$cmd" >/dev/null 2>&1 || exit_error "Missing command: $cmd"
done
```

deploy/docker/gaiahub.sh
criadoperez: fixed.

deploy/docker/gaiahub.sh
criadoperez: Ok. Changed.

deploy/docker/gaiahub.sh
criadoperez: Ok. Changed.

deploy/docker/gaiahub.sh
criadoperez: Ok. Done.

deploy/docker/gaiahub.sh
criadoperez: Yes they are error messages, so directed to stderr. Done.

deploy/docker/gaiahub.sh
criadoperez: Ok. Done.

deploy/docker/gaiahub.sh
```diff
@@ -0,0 +1,104 @@
...
+		echo "GAIA HUB not running."
+		return 0
+	fi
+}
+
+#Starts GAIA HUB
+gh_start() {
+	if check_containers; then
+		echo "GAIA Hub already running. I won't do anything."
+		return
```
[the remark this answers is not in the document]
criadoperez: I don't think so. If a user wants to start a service this is already running, I wouldn't call that an error. Service stays on and informs the user that it already was on.
If you feel differently about this, and prefer to add another message, please add it directly or tell me the message you want and I'll add it.

deploy/docker/gaiahub.sh
criadoperez: Good advice. Added.

jcnelson (approved):
LGTM!

PR #373, opened 2022-04-12
Fix/node dependencies

criadoperez:
Fixes required node dependencies to run tests and solver security dependencies


criadoperez:
Tests run correctly with upgraded dependencies

PR #377, opened 2022-04-13
Migrate from failing CircleCI to Github action

criadoperez:
Migrate from failing CircleCI to Github action


criadoperez:
CircleCI was using node 12 that will reach end of life in 5 days. 2022-04-30.
Current LTS node version is node 16 but is producing some issues, so upgrading instead to node 14 that has reaches end of life on 2023-04-30

